PR #2, opened 2016-12-01
Simplify implementation

norio-nomura:
Also fix off-by-one bug


jpsim:
Fantastic, thanks! Do you need this in a release to address the off-by-one bug?

jpsim:
I've released 0.1.3 in git tags and CocoaPods.

PR #4, opened 2016-12-05
Fix lifetime issue of allocated memory by implicit transformation

norio-nomura:
On passing `String` as `UnsafePointer<UInt8>` arguments to function,
Swift implicitly creates null terminated UTF8 string buffer and passes the pointer to function.
That memory is released immediately on returning the function.
But, yaml_parser requires that is valid while `yaml_parser_t` exists.
This change extends lifetime of the buffer to end of function.

The issue was introduced by my commit 3125c9a. 🙇


jpsim:
Thanks!

Issue #8, opened 2016-12-07
Support Language-Independent Types for YAML™


norio-nomura:
http://yaml.org/type/index.html

norio-nomura:
Some types are not supported by libYAML.

norio-nomura:
This is supported by using `Resolver` and `Constructor`.

Issue #7, opened 2016-12-07
Support explicit type conversion from scalar to Swift types by using TAG


norio-nomura:
It seems `yaml_parser_load()` sets `tag:yaml.org,2002:str` to `yaml_node_s.tag` on both of following:
```yml
---
explicit-string: !!str foo
implicit-string: bar
```
For supporting this over #6, we need to change from using `yaml_parser_load()` to event based parsing.

norio-nomura:
This is supported by using `Resolver` and `Constructor`.

PR #11, opened 2016-12-15
Change to use `yaml_parser_parse()` instead of `yaml_parser_load()`

### LINKED ISSUE #6: Support implicit conversion from scalar to Swift types
e.g.
```swift
extension Node {
    public var any: Any {
        switch self {
        case let .scalar(string):
            if let int = Int(string) {
                return int
            } else if let double = Double(string) {
                return double
            }
            switch string.lowercased() {
            case "true": return true
            case "false": return false
            case "null": return NSNull()
            default: return string
            }
        case let .mapping(dictionary):
            var result = [String:Any](minimumCapacity: dictionary.count)
            dictionary.forEach { result[$0] = $1.any }
            return result
        case let .sequence(array):
            return array.map { $0.any }
        }
    }
}
```

norio-nomura:
This does:
- Add `Parser` class that representing of `yaml_parser_t`
- Add `Tag` ~enum~ class
- Change `Node`
  - Add `Tag` to each case’s associated value
  - ~Change `.mapping` to holding `[Node:Node]`~
  - Remove `init(string:)`
  - Add `array` property that returns `[Node]?`
  - Add `dictionary` property that returns `[Node:Node]?`
  - Add `string` property that returns `String?`

TODOs on this PR:
- [x] Add `Resolver` that resolve `scalar` into Swift Types. Fixes #6, #7 and part of #8.
- [x] Add more test cases
- [x] Add Performance tests

This PR rewrites almost all the codes. Is it better to publish in my own repository instead of PR against Yams?


jpsim:
> This PR rewrites almost all the codes. Is it better to publish in my own repository instead of PR against Yams?

This PR definitely gets Yams closer to where I'd like it to be. I whipped up this project _super_ quickly with the intent to rewrite it later, mostly as a workaround to the performance issues I was seeing in YamlSwift.

I'm also ok if you want to "own" this repo @norio-nomura, I'll happily transfer it over to you if it's something you want to maintain long term. Otherwise, let's address the TODOs (or not, they don't seem critical) and get this in.

I'm also perfectly fine if we keep the repo here, but @norio-nomura no need to wait for my approval or review on things if you're confident.

norio-nomura:
@jpsim Thank you for replying.
> I'm also perfectly fine if we keep the repo here,

I'll continue updating at here. 👍

norio-nomura:
implemented `Resolver` and `Constructor` with reference to PyYAML.

norio-nomura:
Performance test added that simulating process in `SourceKittenFramework.Moudel.init(spmName:)`

Result in `norionomura/sourcekit:302` running on 2.6 GHz Intel Core i5 MacBook Pro:
> Yams/Tests/YamsTests/PerformanceTests.swift:22: Test Case 'PerformanceTests.testSourceKittenIssue289' measured [Time, seconds] average: 0.045, relative standard deviation: 13.045%, values: [0.056026, 0.055852, 0.041780, 0.043248, 0.040798, 0.044501, 0.042547, 0.039861, 0.042329, 0.043930], performanceMetricID:org.swift.XCTPerformanceMetric_WallClockTime, maxPercentRelativeStandardDeviation: 10.000%, maxStandardDeviation: 0.100

Linux version is not extremely slow, it is the same as mac version.
As profiled on mac, half of duration is used in `specialized String.utf8CString.getter`:
screenshot 2016-12-24 15 26 30

norio-nomura:
Global functions are added that encapsulate operation using `Parser` class:
- `Yams.load(yaml:resolver:constructor:)` loads single document and returns root node as `Any`
- `Yams.load_all(yaml:resolver:constructor:)` returns `YamlSequence<Any>` that iterates documents as `Any`
- `Yams.compose(yaml:resolver:constructor:)` loads single document and returns root node as `Node`
- `Yams.compose_all(yaml:resolve:constructor:)` returns `YamlSequence<Node>` that iterates documents as `Node`

Each functions take three parameters:
- YAML `String`
- `Resolver` resolve scalar Node's tag by their contents on generating `Any`. `Resolver.default` is static member that intended to be used as default parameter. Consumer can use their own customized `Resolver` instance that supports their own tags.
- `Constructor` constructs Swift types from `Node`. On constructing `Any`, that uses resolved tag. `Constructor.default` is static member that intended to be used as default parameter. Consumer can use their own customized `Resolver` instance that supports constructing their own types.

Now, `Yams.load*(yaml:)` can generate `Any` that can be used on usual cases.
I want to enhance `Node` that can be used for:
- accessing tree of `Node` easily without constructing `Any`
- emitting YAML string

But, those should be done in another PR.

Could you please review this?

jpsim:
I reviewed #13 instead.

norio-nomura:
Thanks! 🙏

PR #12, opened 2016-12-18
Importing C modules with umbrella directories prevents debugging

RLovelett:
This is similar to a pull request I made against the https://github.com/norio-nomura/Clang_C/pull/1 repository.

The difference is that since you do not already provide an explicit module map (I assume you do not want one) by providing this file SwiftPM will auto generate one.

This work-around was originally suggested by @aciidb0mb3r on the SwiftPM Slack channel.

### DIFF
--- a/Sources/CYaml/include/CYaml.h
+++ b/Sources/CYaml/include/CYaml.h
@@ -1,0 +1,1 @@
+#include "yaml.h"


jpsim:
Awesome, thanks!

PR #13, opened 2016-12-26
Improve `Node` with adding typed accessors

norio-nomura:
This PR based on #11 and against to `master`.

### Add typed accessor properties to `Node`:
- `string`
- `bool`
- `float`
- `null`
- `int`
- `binary`
- `timestamp`

### Convenient accessor method to `Node.sequence`:
- `array()` returns `[Node]`
- `array<Type: ScalarConstructible>()` returns `[Type]`
- `array<Type: ScalarConstructible>(of type: Type.Type)` returns `[Type]`
```swift
let sequence: Node = [
    "true",
    "1.0",
    "1",
    .scalar(base64String, .implicit)
]
XCTAssertEqual(sequence.array(), ["true", "1.0", "1", .scalar(base64String, .implicit)] as [Node])
XCTAssertEqual(sequence.array(of: String.self), ["true", "1.0", "1", base64String])
XCTAssertEqual(sequence.array() as [String], ["true", "1.0", "1", base64String])
XCTAssertEqual(sequence.array(of: Bool.self), [true])
XCTAssertEqual(sequence.array() as [Bool], [true])
XCTAssertEqual(sequence.array(of: Double.self), [1.0, 1.0])
XCTAssertEqual(sequence.array() as [Double], [1.0, 1.0])
XCTAssertEqual(sequence.array(of: Int.self), [1])
XCTAssertEqual(sequence.array() as [Int], [1])
```

### Add  subscript to `Node.mapping`:
```swift
guard let yaml = try Yams.compose(yaml: yamlString),
    let moduleCommand = yaml["commands"]?["<SourceKittenFramework.module>"],
    let imports = moduleCommand["import-paths"]?.array(of: String.self),
    let otherArguments = moduleCommand["other-args"]?.array(of: String.self),
    let sources = moduleCommand["sources"]?.array(of: String.self) else {
        XCTFail("Invalid result form Yams.load()")
        return
}
XCTAssertEqual(imports, self.expectedImports)
XCTAssertEqual(otherArguments, self.expectedOtherArguments)
XCTAssertEqual(sources, self.expectedSources)
```

## Using typed accessors is faster than `[String:Any]` operations because converting operations are minimized.
- `[String:Any]` operations
> Test Case 'PerformanceTests.testSourceKittenIssue289Load' measured [Time, seconds] average: 0.041, relative standard deviation: 11.743%, values: [0.052250, 0.045143, 0.037565, 0.041266, 0.037317, 0.038670, 0.037988, 0.038400, 0.038109, 0.038300], performanceMetricID:org.swift.XCTPerformanceMetric_WallClockTime, maxPercentRelativeStandardDeviation: 10.000%, maxStandardDeviation: 0.100

- Typed accessor operations:
> Test Case 'PerformanceTests.testSourceKittenIssue289Compose' measured [Time, seconds] average: 0.013, relative standard deviation: 25.008%, values: [0.010489, 0.010995, 0.009678, 0.017652, 0.017646, 0.010401, 0.012421, 0.012176, 0.009676, 0.015881], performanceMetricID:org.swift.XCTPerformanceMetric_WallClockTime, maxPercentRelativeStandardDeviation: 10.000%, maxStandardDeviation: 0.100

## Some features are not supported on typed accessor properties:
[!!merge](http://yaml.org/type/merge.html) and [!!value](http://yaml.org/type/value.html) are not supported on using typed accessor properties, because those are handled in constructing dictionary from `Node.mapping`.

Happy Holidays!


Sources/Yams/Tag.swift
```diff
@@ -0,0 +1,120 @@
...
+    public static let int: Tag.Name  = "tag:yaml.org,2002:int"
+    // http://yaml.org/type/index.html
+    public static let binary: Tag.Name  = "tag:yaml.org,2002:binary"
+    public static let merge: Tag.Name  = "tag:yaml.org,2002:merge"
+    public static let omap: Tag.Name  = "tag:yaml.org,2002:omap"
+    public static let pairs: Tag.Name  = "tag:yaml.org,2002:pairs"
+    public static let set: Tag.Name  = "tag:yaml.org,2002:set"
+    public static let timestamp: Tag.Name  = "tag:yaml.org,2002:timestamp"
+    public static let value: Tag.Name  = "tag:yaml.org,2002:value"
+//    public static let yaml: Tag.Name  = "tag:yaml.org,2002:yaml"
```
jpsim: why is this one commented out?

Sources/Yams/Constructor.swift
```diff
@@ -0,0 +1,360 @@
...
+// MARK: - ScalarConstructible
+public protocol ScalarConstructible {
+    // We don't use overloading `init?(_ node: Node)`
+    // because that causes difficulties on using `init` as cosure
+    static func construct(from node: Node) -> Self?
+}
+
+extension Bool: ScalarConstructible {
+    public static func construct(from node: Node) -> Bool? {
+        guard let scalar = node.scalar else { fatalError("Never happen this") }
```
norio-nomura: I'll change all of them to force unwrapping.

Sources/Yams/Constructor.swift
```diff
@@ -0,0 +1,360 @@
...
+        }
+        return .mapping(merge + pairs, node.tag)
+    }
+}
+
+extension Set {
+    public static func construct_set(from node: Node) -> Set<AnyHashable>? {
+        guard let pairs = node.pairs else { fatalError("Never happen this") }
+        // TODO: YAML supports Hashable elements other than str.
+        return Set<AnyHashable>(pairs.map({ String._construct(from: $0.key) as AnyHashable }))
```
norio-nomura: This is required.
This static function is defined in `extension Set {}`.
Swift compiler can't infer the type of generic parameter at here.
```swift
extension Set {
    public static func takes(array: [AnyHashable]) -> Set<AnyHashable> {
        return Set(array)
//    Playground execution failed: error: 104351.xcplaygroundpage:3:16:
//    error: expression type 'Set<Element>' is ambiguous without more context
//    return Set(array)
    }
}
```

Sources/Yams/Constructor.swift
```diff
@@ -0,0 +1,360 @@
...
+            }
+            if let minuteInSecond = components[10].flatMap({ Int($0) }).map({ $0 * 60 }) {
+                seconds += minuteInSecond
+            }
+            if components[8] == "-" { // sign
+                seconds *= -1
+            }
+            return TimeZone(secondsFromGMT: seconds)
+        }()
+        // Using `DateComponents.date` causes crash on Linux
```
norio-nomura: Should we report crash on `NSUnimplemented()`?
```
Welcome to Swift version 3.0.2 (swift-3.0.2-RELEASE). Type :help for assistance.
  1> import Foundation
  2> DateComponents(calendar: Calendar(identifier: .gregorian), year: 2016).date 
fatal error: copy(with:) is not yet implemented: file Foundation/NSCalendar.swift, line 1434
Execution interrupted. Enter code to recover and continue.
Enter LLDB commands to investigate (type :help for assistance.)
  3>  
```

Sources/Yams/Constructor.swift
jpsim: Ok!

Sources/Yams/Constructor.swift
jpsim: Oh, no, that's fine. Sorry I've been reviewing on my phone so I didn't look it up.

norio-nomura: I'll change the comment. 👍

Sources/Yams/Parser.swift
```diff
@@ -0,0 +1,315 @@
...
+        yaml = string
+        self.resolver = resolver
+        self.constructor = constructor
+
+        yaml_parser_initialize(&parser)
+#if USE_UTF16
+        yaml_parser_set_encoding(&parser, YAML_UTF16BE_ENCODING)
+        data = yaml.data(using: .utf16BigEndian)!
+        data.withUnsafeBytes { bytes in
+            yaml_parser_set_input_string(&parser, bytes, data.count)
```
norio-nomura: Generated data does not contain null termination.

Sources/Yams/Tag.swift
norio-nomura: I could not decide how we handle this.
PyYAML is not implemented to handling "tag:yaml.org,2002:yaml".
When I read [the specification](http://yaml.org/type/yaml.html) again, it seems that it can not be implemented.
I'll change that comment to declaration and add comment describing we don't support that.

Sources/Yams/Constructor.swift
```diff
@@ -0,0 +1,360 @@
...
+        guard var scalar = node.scalar else { fatalError("Never happen this") }
+        switch scalar {
+        case ".inf", ".Inf", ".INF", "+.inf", "+.Inf", "+.INF":
+            return Double.infinity
+        case "-.inf", "-.Inf", "-.INF":
+            return -Double.infinity
+        case ".nan", ".NaN", ".NAN":
+            return Double.nan
+        default:
+            scalar = scalar.replacingOccurrences(of: "_", with: "")
```
norio-nomura: I'll fix this for today.

norio-nomura:
Thanks for reviews!
Updated.

Sources/Yams/Parser.swift
jpsim: :+1:

Sources/Yams/Constructor.swift
norio-nomura: I think this is addressed on 07de411.

jpsim:
Any reason not to merge this? It looks great!

norio-nomura:
Sorry for delay. 🙇 
I thought #15 would break some APIs.
So, I think this would be merged to master via #15.

jpsim:
Ok, so you're saying you'd rather not merge this but instead finish up #15?

norio-nomura:
> Ok, so you're saying you'd rather not merge this but instead finish up #15?

Ah, Sorry my bad.
I didn't notice that complicates reviewing.
Let's release `0.1.5` by #17 based on current master and merge this to master after that.

norio-nomura:
After merging #17, I'll resolve conflicts on this and merge to master.

norio-nomura:
Conflicts has been fixed.
Waiting CI.

PR #17, opened 2017-01-26
Release 0.1.5

norio-nomura:
Release notes would be:
- Importing C modules with umbrella directories prevents debugging
- Fix “include of non-modular header inside framework module 'Yams’” error on Xcode 8.3
- Support iOS, tvOS and watchOS


.travis.yml
```diff
@@ -2,18 +2,23 @@ language: generic
 matrix:
   include:
     - script:
-        - xcodebuild -scheme Yams test
+        - xcodebuild -scheme Yams build-for-testing test-without-building # -sdk macosx -destination arch=x86_64
```
jpsim: why `build-for-testing test-without-building` rather than `test`?

norio-nomura: That was workaround for unstable testing on Xcode 8.
I don't know whether the workaround is still needed or not anymore.
https://github.com/Quick/Quick/commit/ed9c17b8c39a5a36d0244a2e1ff4450a4b971a30

norio-nomura:
> Just a question, shouldn't block merging if there's a good reason for it.

You're right. Sorry for confusion.
I'll merge another PR today that approved and I'm blocking.

PR #21, opened 2017-02-23
Change to use UTF16 by default

norio-nomura:
Because using UTF16 is faster than using UTF8 on macOS:

|                                      |UTF8 |UTF16|
|--------------------------------------|-----|-----|
|`testSourceKittenIssue289UsingCompose`|0.026|0.007|
|`testSourceKittenIssue289UsingLoad`   |0.038|0.017|

The `USE_UTF8` option using UTF8 will remain.


jpsim:
👍

norio-nomura:
🙏

Issue #20, opened 2017-02-05
[question] Is it a deliberate design choice that Pair's key & value are not public?


RLovelett:
It seems that [`Pair`'s `key` & `value` are not `public`](https://github.com/jpsim/Yams/blob/8551f6d0841e28b31ba5f793bed11f215295c54d/Sources/Yams/Node.swift#L122-L123) which makes this inaccessible outside the module. Is this a deliberate design choice?

If so how is someone supposed to find the value of a `Node`?

norio-nomura:
I've made them `internal` as implementation details. Since direct access to them may result in operations out of the specification of YAML.
In the latest master, the `mapping` property behaves like` Dictionary`:
https://github.com/jpsim/Yams/blob/master/Tests/YamsTests/NodeTests.swift#L126-L156
I guess this is enough, but would you please tell me if there is a reason to need direct access that can not be covered by these?

norio-nomura:
First I wondered whether `Node.mapping` should have`[Pair<Node>]` or `Dictionary<Node, Node>`, but respecting the very first implementation `[Pair<Node>] `Was selected.
Now I am thinking that it would be better to change to `Dictionary<Node, Node>` as a result of implementing the whole story.

RLovelett:
Sorry for the delayed response. Had to step away for this for awhile.

Ultimately I cannot give a really good explanation for what I'm trying to do yet because I'm still trying to figure out what I want to do. Therefore it's unfair of me to make requests of you that I cannot articulate myself. I'm going to close this, revert to 0.1.4 (where I know my code worked); once things stabilize I'll check out PR #22 and the changes you've made on master.

Thank you for the thoughtful responses.

PR #22, opened 2017-02-23
Change associated value of `Node`

norio-nomura:
- Changes associated value of `Node`:
  - `Node.scalar(String, Tag, Scalar.Style)` to `Node.scalar(Scalar)`
  - `Node.mapping([Pair<Node>], Tag, Mapping.Style)` to `Node.mapping(Mapping)`
  - `Node.sequence([Node], Tag, Sequence.Style)` to `Node.sequence(Sequence)`

- Add initializers to `Node`:
  - `Node.init(_ pairs: [(Node, Node)], _ tag: Tag = .implicit, _ style: Mapping.Style = .any)`
  - `Node.init(_ nodes: [Node], _ tag: Tag = .implicit, _ style: Sequence.Style = .any)`

- Access level of `Pair` is changed to `internal`

I expect this would be an answer to #20.

### DIFF
--- a/Sources/Yams/Constructor.swift
+++ b/Sources/Yams/Constructor.swift
@@ -258,8 +258,8 @@ extension Dictionary {
     private static func flatten_mapping(_ node: Node) -> Node {
         assert(node.isMapping) // swiftlint:disable:next force_unwrapping
         let mapping = node.mapping!
-        var pairs = mapping.pairs
-        var merge = [Pair<Node>]()
+        var pairs = mapping.pairs.map(Pair.toTuple)
+        var merge = [(key: Node, value: Node)]()
         var index = pairs.startIndex
         while index < pairs.count {
             let pair = pairs[index]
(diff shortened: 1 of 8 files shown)


norio-nomura:
`Node.init(_:_:_:)` are three overloaded initializers.
```swift
extension Node {
    public init(_ string: String, _ tag: Tag = .implicit, _ style: Scalar.Style = .any) {
        self = .scalar(.init(string, tag, style))
    }

    public init(_ pairs: [(Node, Node)], _ tag: Tag = .implicit, _ style: Mapping.Style = .any) {
            self = .mapping(.init(pairs, tag, style))
    }

    public init(_ nodes: [Node], _ tag: Tag = .implicit, _ style: Sequence.Style = .any) {
        self = .sequence(.init(nodes, tag, style))
    }
}
```
I was wondering whether to use the external label of the first parameter, do not use now.

PR #28, opened 2017-03-27
Disable failing SwiftLint rule.

### LINKED ISSUE #27: Linting fails with SwiftLint v. 0.17.0
SwiftLint v. 0.17.0 has a new rule `identifier_name` which fails in `Yams/Tests/YamsTests/TestHelper.swift:46:2`. Running the linter is a run script in the build phases and because of that it's not possible to build the project. Funnily enough this also breaks Carthage builds and therefore building SwiftLint itself, which has Yams as one of its dependencies (via SourceKitten) 😃

Lukas-Stuehrk:
This fixes #27.


norio-nomura:
Thanks! 🙏

PR #36, opened 2017-04-11
Make Node locatable in YAML

norio-nomura:
This change provides the location information of the Node in YAML to the consumer.
Consumers can use the location information of Node to display detailed messages for the contents of YAML. (e.g. locatable deprecation warning in `.swiftlint.yml` for Xcode)

- Change `YamlError.Mark` to `Mark`
- Add `Mark` to `Node`

Can we treat this as patch level changes?


jpsim:
> Can we treat this as patch level changes?

Sure.

norio-nomura:
Thanks! 🙏

PR #37, opened 2017-04-14
Fix crash if context is nil in `YAML_SCANNER_ERROR`

norio-nomura:
Fixes https://github.com/realm/SwiftLint/issues/1436


norio-nomura:
🙏

PR #45, opened 2017-06-16
Fix static analyzer issues

norio-nomura:


jpsim:
Can't we just remove the dummy vars and do `(void)POP(parser, parser->simple_keys);`?

PR #46, opened 2017-06-24
Update to compile with Xcode 9 beta 2

jpsim:


jpsim:
I agree, I mostly just want to avoid Xcode's prompt to update. Should still avoid it with `SWIFT_VERSION = 3.0;` though.

norio-nomura:
Thanks!

PR #50, opened 2017-07-28
[travis] Override SWIFT_VERSION=4.0 on JOB=Xcode9 SWIFT_VERSION=4.0

norio-nomura:


norio-nomura:
confirmed that `-swift-version 4` was used on [travis's job](https://travis-ci.org/jpsim/Yams/jobs/258434075#L85).

jpsim:
Doesn't the environment variable that's already set do the same thing?

norio-nomura:
No.
[The build log](https://travis-ci.org/jpsim/Yams/jobs/258288660) of #49 does not contain `-swift-version 4`.

norio-nomura:
I am sorry that my English is not good.
In order to override the build settings when running `xcodebuild`, we need to pass `SWIFT_VERSION=4.0` as a command line option. In #49 you can see from [the log](https://travis-ci.org/jpsim/Yams/jobs/258288660) that `-swift-version 3` is used.

jpsim:
ok, thanks!

PR #52, opened 2017-08-11
Use `..<` instead of `...` as equivalent to `substring(to:)`

norio-nomura:
Fix https://travis-ci.org/jpsim/Yams/jobs/263299535#L102
> Test Case '-[YamsTests.EncoderTests testEncodingDate]' started.
> fatal error: cannot increment beyond endIndex

### DIFF
--- a/Tests/YamsTests/ConstructorTests.swift
+++ b/Tests/YamsTests/ConstructorTests.swift
@@ -351,6 +351,23 @@ class ConstructorTests: XCTestCase { // swiftlint:disable:this type_body_length
         YamsAssertEqual(objects, expected)
     }
 
+    func testTimestampWithNanosecond() throws {
+        #if os(Linux)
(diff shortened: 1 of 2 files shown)


Tests/YamsTests/ConstructorTests.swift
```diff
@@ -351,6 +351,23 @@ class ConstructorTests: XCTestCase { // swiftlint:disable:this type_body_length
         YamsAssertEqual(objects, expected)
     }
 
+    func testTimestampWithNanosecond() throws {
+        #if os(Linux)
+            // FIXME: swift-corelibs-foundation can't format date with nanosecond.
+            // https://bugs.swift.org/browse/SR-3158
+        #else
+            let example = [
```
jpsim: Why not just `let example = "nanosecond: 2001-12-15T02:59:43.123456789Z\n"`?

norio-nomura:
No problem. 🙂
Thanks for review!

Issue #55, opened 2017-08-31
Change the `Emitter` and` Parser` options to properties.


norio-nomura:

jpsim:
Done in #57.

PR #43, opened 2017-05-07
Support SE-0166 Swift Archival & Serialization

norio-nomura:
Support [SE-0166 Swift Archival & Serialization](https://github.com/apple/swift-evolution/blob/master/proposals/0166-swift-archival-serialization.md)
Fixes #38 

~This uses local built toolchain https://github.com/norio-nomura/swift-dev/releases/tag/20170501a~

TODOs for merging:
- ~~[ ] Add text encoding detection to `YAMLDecoder.decode(_:from:)`~~
- [x] Revise throwing errors on failing Encode and Decode
- [x] Add more tests
- [x] Support first snapshot of Swift 4.0

PR #57, opened 2017-10-20
Add `Emitter.Options`

norio-nomura:
#55

### DIFF
--- a/Sources/Yams/Emitter.swift
+++ b/Sources/Yams/Emitter.swift
@@ -195,7 +195,7 @@ public final class Emitter {
         /// Set the preferred line break.
         public var lineBreak: LineBreak = .ln
 
-        // internal since we don't know these should be exposed.
+        // internal since we don't know if these should be exposed.
         var explicitStart: Bool = false
         var explicitEnd: Bool = false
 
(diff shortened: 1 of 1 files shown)


norio-nomura:
This will be needed by `YAMLEncoder` in #43

Sources/Yams/Emitter.swift
```diff
@@ -183,9 +183,31 @@ public final class Emitter {
...
+        /// Set the intendation increment.
+        public var indent: Int = 0
+        /// Set the preferred line width. -1 means unlimited.
+        public var width: Int = 0
+        /// Set if unescaped non-ASCII characters are allowed.
+        public var allowUnicode: Bool = false
+        /// Set the preferred line break.
+        public var lineBreak: LineBreak = .ln
+
+        // internal since we don't know these should be exposed.
```
jpsim: missing a word here: "internal since we don't know if these should be exposed." or maybe "how these should be exposed." depending on what you mean.

norio-nomura: Thanks!
I don't mean *how*, maybe *if*.

PR #59, opened 2017-10-25
Add negative Int support to other formats in addition to canonical

norio-nomura:


norio-nomura:
#43 depends on this change.

PR #61, opened 2017-10-26
Add `Emitter.Options.sortKeys`

norio-nomura:
Set if emitter should sort keys in lexicographic order.
This will be used for stabilizing `EncoderTests.testEncodingTopLevelStructuredSingleClass` in #43 on Linux.

### DIFF
--- a/Tests/YamsTests/EmitterTests.swift
+++ b/Tests/YamsTests/EmitterTests.swift
@@ -110,7 +110,7 @@ class EmitterTests: XCTestCase {
         }
     }
 
-    func testSortedKeys() throws {
+    func testSortKeys() throws {
         let node: Node = [
             "key3": "value3",
             "key2": "value2",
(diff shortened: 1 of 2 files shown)


Tests/YamsTests/EmitterTests.swift
```diff
@@ -109,6 +109,20 @@ class EmitterTests: XCTestCase {
...
+
+    func testSortedKeys() throws {
+        let node: Node = [
+            "key3": "value3",
+            "key2": "value2",
+            "key1": "value1"
+        ]
+        let yaml = try Yams.serialize(node: node)
+        let expected = "key3: value3\nkey2: value2\nkey1: value1\n"
+        XCTAssertEqual(yaml, expected)
```
jpsim: Will this be flaky on Linux like `testEncodingTopLevelStructuredSingleClass` is right now because the order isn't guaranteed?

norio-nomura: This won't be flaky on Linux, because this test does not use `Swift.Dictionary` that causing issue on Linux.

norio-nomura:
Thank you for review!
Renamed to `sortKeys`. 👍

PR #67, opened 2017-11-01
Use `Resolver.basic` in `YAMLDecoder.decode()`

norio-nomura:
Since `YAMLDecoder` does not need to resolve type,` Resolver.basic` which does not resolve type is enough.


norio-nomura:
I think that we can start dropping swift <3.2 when #66 and this will be merged.

PR #68, opened 2017-11-01
Refactor String.sexagesimal()

jpsim:
* to use single initialization
* to reduce protocol requirements
* to use Sequence.reduce() rather than forEach with two vars


jpsim:
Main thing I'd like from a review here is if the protocol requirement changes are acceptable, given that it's a public protocol.

PR #69, opened 2017-11-02
Require Swift 3.2 or later to build

jpsim:


jpsim:
Better reviewed with whitespace turned off: https://github.com/jpsim/Yams/pull/69/files?w=1

PR #79, opened 2017-11-03
Add performance test using Decodable

norio-nomura:
Performance is good in order of using `compose()`, using `Swift.Decodable`, using `load()`.


jpsim:
Nice.

PR #82, opened 2017-11-04
Use Multiline String Literals

norio-nomura:


norio-nomura:
Would it be unnecessary to test using `SWIFT_VERSION = 3.2`?

norio-nomura:
Because there is no `swift-3.2-RELEASE` compiler, therefore "Swift>=3.2 is required" has the same meaning as "Swift>=4.0 is required".
In other words, if we do not support Swift 3.1.1, it seems pointless to say that we support Swift 3.2.

jpsim:
Swift 3.2 is still a swift version, distinct from Swift 4.x, even if it's technically compiled using the same compiler as Swift 4.0.

norio-nomura:
Hmm, I understand.

norio-nomura:
Thanks for review! 🙏

PR #88, opened 2017-12-01
Add `pod lib lint`

norio-nomura:


norio-nomura:
I'm not sure too. However, as far as we say it supports Swift 3.2 or later, I think we should test them as well.

Issue #73, opened 2017-11-02
Add usage docs in markdown


jpsim:
Could also be rendered with the API docs (#72) using Jazzy's "guides" functionality.

lynchsft:
Issue resolved in #91 and improved in many subsequent PRs.

PR #91, opened 2017-12-08
Add Usage section to README.md

norio-nomura:
#73

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,24 +30,22 @@ Add `github "jpsim/Yams"` to your `Cartfile`.
 
 ## Usage
 
-Each of input / output YAML APIs can be classified into three, depending on the model types to be handled.
+Yams has three groups of conversion APIs:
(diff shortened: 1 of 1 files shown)


norio-nomura:
Should we test with CI that the sample code is buildable?

jpsim:
It would be nice to have a high-level explanation of when to use `serialize` vs `dump` and `load` vs `compose`.

norio-nomura:
Thanks for suggestion! Updated.

jpsim:
> Should we test with CI that the sample code is buildable?

Nice to have, but not necessary for now.

I hope you don't mind, I made some edits. Let me know if you disagree with any of my changes.

Otherwise, feel free to merge this when you're happy with it! 👍

norio-nomura:
I think that it became very good by your editing! Thanks! 🙏

Issue #84, opened 2017-11-06
Add key encoding strategy to YAMLEncoder


norio-nomura:
See: https://github.com/apple/swift/pull/12779

How should `YAMLEncoder` options be?
A. Rename `options` to `emitterOptions`, then add `keyEncodingStrategy` to `YAMLEncoder`
B. Add `YAMLEncoder.options.keyEncodingStrategy`

jpsim:
I think I prefer B.

Also, `convertToSnakeCase`/`convertFromSnakeCase` could be just as useful for YAML as it is for JSON.

lynchsft:
I strongly encourage the use of [MetaCodable library](https://github.com/SwiftyLab/MetaCodable/blob/main/README.md) to handle formatting variations. Will close this (ancient) issue in 90 day sir no objection.

PR #92, opened 2017-12-19
[WIP] Add Key coding strategy

norio-nomura:
#84

This uses different algorithm on `KeyDecodingStrategy` from https://github.com/apple/swift/pull/12779 that exprimenting to fix [SR-6629](https://bugs.swift.org/browse/SR-6629).


norio-nomura:
I will replace this with another PR.

PR #95, opened 2018-01-15
Prevent crash when decoding case of malformed sequence of strings in a mapping

dcutting:
I had a case of malformed input that crashed the Decoder when using a `Codable` struct.

Given a struct like this:

```
struct Sample: Codable {
  let values: [String]
}
```

data like this would decode correctly:

```
values:
- hello
```

but data like this would crash:

```
values:
- hello:
```

It seems to be because it's looking for a mapping but not finding one and that case is not handled, so the subsequent assertion fails. This PR avoids the crash by returning nil instead.

I'm not sure this is the right way to handle it (or even a good way), but at least it prevents the crash in my use case.

Thanks for creating this pod!

### DIFF
--- a/Sources/Yams/Constructor.swift
+++ b/Sources/Yams/Constructor.swift
@@ -21,7 +21,7 @@ public final class Constructor {
         }
         switch node {
         case .scalar:
-            return String._construct(from: node)
+            return String._construct(from: node)! // swiftlint:disable:this force_unwrapping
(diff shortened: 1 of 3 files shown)


norio-nomura:
Thank you for PR! But we already received another PR #94 today that fixes the same issue.
So close this.
Thank you again! 🙏

PR #96, opened 2018-01-26
Support building with Swift 4.1

norio-nomura:
tested with `swift-4.1-DEVELOPMENT-SNAPSHOT-2018-01-25-a` and Xcode 9.3 beta 1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,11 @@ aliases:
     - store_test_results:
         path: build/reports/
 
+  - &steps-for-linux-without-swiftlint
+    - checkout
+    - run: swift test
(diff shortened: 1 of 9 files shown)


norio-nomura:
filed https://github.com/realm/SwiftLint/issues/2027 for `identifier_name` rule violations.

jpsim:
Why did you add functions on `Array` and `LazyMapCollection` as well as `Sequence`? Since `Array` and `LazyMapCollection` both conform to `Sequence`, they should be getting it "for free"?

Is it to force specializations for use in Yams to avoid runtime generics overhead? Because that shouldn't be necessary within a module 🤔

PR #98, opened 2018-01-27
Change access level of `compactMap()` from `public` to `internal`

norio-nomura:
It will avoid interfering with the `compactMap` definition in the dependent packages such as SourceKitten.

### DIFF
--- a/Sources/Yams/shim.swift
+++ b/Sources/Yams/shim.swift
@@ -8,18 +8,6 @@
 
 #if (!swift(>=4.1) && swift(>=4.0)) || !swift(>=3.3)
 
-    extension Array {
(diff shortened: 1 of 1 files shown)


Sources/Yams/shim.swift
```diff
@@ -6,22 +6,22 @@
 //  Copyright (c) 2018 Yams. All rights reserved.
 //
 
-#if !swift(>=4.1)
+#if (!swift(>=4.1) && swift(>=4.0)) || !swift(>=3.3)
```
jpsim: Ugh 🤦‍♂️

norio-nomura:
Thanks for review!
Removed `compactMap` shims from `Array` and `LazyMapCollection`, since I confirmed that those are provided by `Sequence`.

PR #105, opened 2018-02-17
Use more effectively that `Node` is an enum

norio-nomura:
* Change parameter type of `ScalarConstructible.construct(from:)` from `Node` to `Node.Scalar`
* Change `Constructor`:
  * Split `Map` type into `ScalarMap`, `MappingMap` and `SequenceMap`
  * Split `defaultMap` static property into `defaultScalarMap`, `defaultMappingMap` and `defaultSequenceMap`
  * Change `init(_:)` to `init(_:_:_:)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ##### Breaking
 
-* Some APIs are changed related to `ScalarConstructible`
+* Some APIs have changed related to `ScalarConstructible`
   * Change parameter type of `ScalarConstructible.construct(from:)` from `Node`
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,17 @@
 
 ##### Breaking
 
-* None.
+* Some APIs are changed related to `ScalarConstructible`
```
jpsim: are -> have

norio-nomura: Thanks!

norio-nomura:
Thanks for reviews!

PR #107, opened 2018-03-01
Use the concrete type parameter passed to `YAMLDecoder.decode()`

norio-nomura:
Same as https://github.com/apple/swift/pull/12844

### DIFF
--- a/Tests/YamsTests/EncoderTests.swift
+++ b/Tests/YamsTests/EncoderTests.swift
@@ -294,14 +294,14 @@ class EncoderTests: XCTestCase {
 
     func testDecodingConcreteTypeParameter() {
         let encoder = YAMLEncoder()
-        guard let json = try? encoder.encode(Employee.testValue) else {
+        guard let yaml = try? encoder.encode(Employee.testValue) else {
             expectUnreachable("Unable to encode Employee.")
             return
         }
 
         let decoder = YAMLDecoder()
-        guard let decoded = try? decoder.decode(Employee.self as Person.Type, from: json) else {
-            expectUnreachable("Failed to decode Employee as Person from JSON.")
+        guard let decoded = try? decoder.decode(Employee.self as Person.Type, from: yaml) else {
+            expectUnreachable("Failed to decode Employee as Person from YAML.")
             return
         }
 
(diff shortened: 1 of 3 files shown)


norio-nomura:
`pod lib lint` failed with following error:
```console.log
 -> Yams (0.5.0)
    - WARN  | url: There was a problem validating the URL https://github.com/jpsim/Yams.

[!] Yams did not pass validation, due to 1 warning (but you can use `--allow-warnings` to ignore it) and all results apply only to public specs, but you can use `--private` to ignore them if linting the specification for a private pod.
You can use the `--no-clean` option to inspect any issue.
Exited with code 1
```
It can not be reproduced on my local site.
I'll re-run those ci jobs later.

norio-nomura:
The url validation issue on CricleCI caused by ruby's version is 2.0.
Using ruby 2.3 will resolve the issue. 🤔 
```terminal.sh-session
$ /usr/local/lib/ruby/gems/2.3.0/bin/pod lib lint
…
** BUILD SUCCEEDED **

   Testing with xcodebuild. 
 -> Yams (0.5.0)

Yams passed validation.
```

norio-nomura:
`pod lib lint` issue has been resolved on #109.

Tests/YamsTests/EncoderTests.swift
```diff
@@ -292,6 +292,22 @@ class EncoderTests: XCTestCase {
         }
     }
 
+    func testDecodingConcreteTypeParameter() {
+        let encoder = YAMLEncoder()
+        guard let json = try? encoder.encode(Employee.testValue) else {
```
jpsim: Shouldn't the variable be named `yaml` rather than `json`?

Tests/YamsTests/EncoderTests.swift
```diff
@@ -292,6 +292,22 @@ class EncoderTests: XCTestCase {
...
+            expectUnreachable("Unable to encode Employee.")
+            return
+        }
+
+        let decoder = YAMLDecoder()
+        guard let decoded = try? decoder.decode(Employee.self as Person.Type, from: json) else {
+            expectUnreachable("Failed to decode Employee as Person from JSON.")
```
jpsim: "from YAML."?

jpsim:
Will cut a release after merging this.

PR #108, opened 2018-03-02
Use ruby from homebrew

norio-nomura:
SeeAlso: https://github.com/realm/SwiftLint/pull/2068
This will fix failing issue on https://github.com/jpsim/Yams/pull/107


norio-nomura:
Close via #109

PR #111, opened 2018-03-16
Drop support building with `-swift-version 3`

norio-nomura:


norio-nomura:
In accordance with the review by `SE-0169`, the access control of some internal types has also been modified.

PR #113, opened 2018-03-22
Refactor Constructor

norio-nomura:


Sources/Yams/Constructor.swift
```diff
@@ -360,13 +360,34 @@ private extension String {
     }
 }
 
+#if swift(>=4.1)
+
+private extension StringProtocol where Self.SubSequence == Substring {
```
jpsim: Any reason why this can't be

```swift
private extension StringProtocol {
    func substring(from offset: IndexDistance) -> SubSequence {
        return self[index(startIndex, offsetBy: offset)...]
    }
}
```

It looks like it compiles going back to Swift 4.0.0 on both Linux & Darwin, so we wouldn't need the explicit `String` and `Substring` versions below.

Sources/Yams/Constructor.swift
```diff
@@ -128,7 +128,7 @@ extension Date: ScalarConstructible {
             let length = $0.count
             let nanosecond: Int?
             if length < 9 {
-                nanosecond = Int($0 + String(repeating: "0", count: 9 - length))
+                nanosecond = Int($0).map { repeatElement(10, count: 9 - length).reduce($0, *) }
```
jpsim: I'm sorry I don't understand what this is doing and how this patch changes behavior. Could you explain please?

jpsim: I think part of what contributes to my confusion is the shorthand closure param identifier `$0` being used in a pretty complex closure with nested closures.

norio-nomura:
Thanks! 🙏

PR #117, opened 2018-03-27
Fix parse quoted values as strings

### LINKED ISSUE #116: Parse quoted values as strings
Is it possible to configure Yams to parse a quoted value (single or double quotes) as a string, instead of trying to convert to an assumed Swift type?
```yams
version: "10.10" // want it to parse to "10.10" String instead of 10.1 Float
```
I know it's possible to use an explicit type tag as a workaround, but it would be nice to be able to just quote a value which I thought is the standard way of specifying strings in yaml.
```
version: !!str '10.10' // correctly parses as "10.10"
```
This would fix https://github.com/yonaskolb/XcodeGen/issues/268

norio-nomura:
Fix #116

### DIFF
--- a/Sources/Yams/Parser.swift
+++ b/Sources/Yams/Parser.swift
@@ -287,9 +287,8 @@ private class Event {
         return Node.Scalar.Style(rawValue: event.data.scalar.style.rawValue)!
     }
     var scalarTag: String? {
-        guard event.data.scalar.plain_implicit == 0,
-            event.data.scalar.quoted_implicit == 0 else {
-                return nil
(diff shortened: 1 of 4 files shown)


Sources/Yams/Parser.swift
```diff
@@ -287,9 +287,8 @@ private class Event {
         return Node.Scalar.Style(rawValue: event.data.scalar.style.rawValue)!
     }
     var scalarTag: String? {
```
keith: This can return a `String` now

jpsim: It can't since `string(from:)` still returns `String?`

PR #127, opened 2018-06-03
[CircleCI] Change jazzy job to using docker

norio-nomura:


norio-nomura:
This change reduces the duration of  `jazzy` job from [2:03](https://circleci.com/gh/jpsim/Yams/1361) to [0:59](https://circleci.com/gh/jpsim/Yams/1372).
I expect this will be about 0:39, when docker's layer caching will works on all CircleCI's docker servers.

jpsim:
How is jazzy installed on that Docker image? It looks like the job doesn’t use the checked in Gemfile?

norio-nomura:
jazzy has been installed by `gem install jazzy`.
The image's Dockerfile and build log is following: https://hub.docker.com/r/norionomura/sourcekitten/builds/bpxsjafdsjura4rptfjnrmt/

jpsim:
Can’t we keep the bundler installation instead?

norio-nomura:
I think that we can. What is the advantage of installing with bundler?

norio-nomura:
Hm, maybe I should make `norionomura/jazzy:0.9.3_swift-4.1.2` instead of `norionomura/sourcekitten:0.21.0_swift-4.1.2` for enabling pinning the version of jazzy. 🤔

norio-nomura:
I have created [norionomura/jazzy](https://hub.docker.com/r/norionomura/jazzy/) and changed to use that with pinning jazzy version to 0.9.3 by image tag.

jpsim:
Ok, but I don't think that was necessary, sorry for the delay in responding. My concern was that the version specified in Gemfile.lock didn't correspond to the version we're running on CI, which isn't actually resolved by pinning the Docker image to a jazzy version.

In any case, that's a small concern. I'm ok with either the `norionomura/jazzy` image or the one used before `norionomura/sourcekitten`. Up to you!

👍

norio-nomura:
> My concern was that the version specified in Gemfile.lock didn't correspond to the version we're running on CI, which isn't actually resolved by pinning the Docker image to a jazzy version.

Thanks for explanation! 
:thinking: I'll try to use bundler with my docker image.

norio-nomura:
I found another caching issues and opened #128 for fixing them.

jpsim:
Presumably with #128 we don't need to use the jazzy docker image?

norio-nomura:
The CircleCI's OSS plan supports parallel execution of Linux jobs, but does not support parallel execution of macOS jobs. So, I think that it is worth changing from macOS to docker.

jpsim:
Oh yes, I'm still in favor of using Docker, I meant we don't need Jazzy pre-installed in a Docker image if we're installing it via Bundler. i.e. change `norionomura/jazzy:0.9.3_swift-4.1.2` to `norionomura/sourcekitten:swift-4.1.2`?

norio-nomura:
Oh, I misunderstood.
I see. However, I have already deleted `norionomura/sourcekitten` from the docker hub and I think that contents will be the same as the jazzy image even if I make it again. So I want to proceed as it is.

norio-nomura:
Thanks! 🙏

PR #132, opened 2018-07-20
Set `DYLIB_COMPATIBILITY_VERSION` and `DYLIB_CURRENT_VERSION` in `Yams.xcodeproj`

### LINKED ISSUE #131: When built with Carthage, version is 0.0.0
I've been trying to use the Stylist framework which has a dependency on Yams. Stylist is expecting version 1.0.0 of Yams, but when the apps tries to start, it returns this message:

dyld: Library not loaded: @rpath/Yams.framework/Yams
  Referenced from: /Users/derekclarkson/Library/Developer/CoreSimulator/Devices/5951E8ED-55B2-4E70-BDE3-0858F47C7C40/data/Containers/Bundle/Application/9FD14F8C-C4D8-4735-AAC0-46D5C61F74A4/Crux.app/Frameworks/Stylist.framework/Stylist
  Reason: Incompatible library version: Stylist requires version 1.0.0 or later, but Yams provides version 0.0.0

I've no idea at this time why Yams is reporting 0.0.0 as a version.

norio-nomura:
Since CocoaPods installs dependencies with setting `DYLIB_*_VERSION` as `1`, `DYLIB_*_VERSION` did not match between `Yams.framework` built with own `Yams.xcodeproj` and installed by CocoaPods.
This change matches them and would resolve #131.


norio-nomura:
These version numbers are used for binary compatibility by dyld.
There is no reason to update them until Swift provides ABI stability and we start binary distribution.

jpsim:
Got it 👍

PR #138, opened 2018-12-04
Update CircleCI to 2.1

norio-nomura:


.circleci/config.yml
```diff
@@ -181,4 +154,6 @@ workflows:
               only:
                 - master
     jobs:
-      - linux_swift_4.2
+      - linux_on_docker:
+          name: linux_swift_4.2
+          image: norionomura/swift:swift-4.2-branch
```
norio-nomura: I'll open another PR updating daily to `swift-5.0-branch`, since it needs workarounds for crashing test on `swift-5.0-branch`.

jpsim:
Sorry it looks like I left the Azure Pipelines enabled for the whole project, even though I was just experimenting in #137. I disabled it and force-pushed the same last commit here to trigger another CI pass.

.circleci/config.yml
```diff
@@ -1,176 +1,149 @@
...
-      - xcode_9.4_swift_4.1.2
-      - xcode_10_swift_4.2
+      - xcode:
+          name: xcode_9.2_swift_4.0.3
+          xcode: "9.2.0"
+          post-steps: [pod_lib_lint]
+      - xcode:
+          name: xcode_9.4_swift_4.1.2
+          xcode: "9.4.1"
+          post-steps: [pod_lib_lint]
```
jpsim: If we run pod lib lint on anything, shouldn't it be the latest Xcode version?

norio-nomura: I think that the old version is sufficient until cocoapods supporting Xcode 10 is released.
Or should we test with beta version of cocoapods?

jpsim: Oh right, that's why we're not doing this on Xcode 10, because CocoaPods stable doesn't yet support it. Good for now then! 👍

norio-nomura:
By the way, it seems that multiple macOS jobs run in parallel on CircleCI, did you upgrade your CircleCI plan?

jpsim:
> By the way, it seems that multiple macOS jobs run in parallel on CircleCI, did you upgrade your CircleCI plan?

Yes, sort of.

norio-nomura:
Thank you very much. 🙏

PR #145, opened 2018-12-30
Enhance encoding support

norio-nomura:
- [x] Switch `Parser`'s preferred encoding on runtime
- [ ] ~~Switch `Emitter`'s preferred encoding on runtime~~
- [x] Optimize performance on Swift 5.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ##### Breaking
 
-* None.
+* Change `byteOffset` to `offset` in `YamlError.reader`.  
+  [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,12 @@
 
 ##### Enhancements
 
-* None.
+* Add `encoding` option to `Parser` as `Parser.Encoding` type for select
+  encoding on calling CYaml API. Along with that change, `encoding` option
+  are added to `load()`, `load_all()`, `compose()`, `compose_all` and 
+  `YAMLDecoder`. Default value of `encoding` option will be selected by
+  runtime environment.   
+  [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,11 @@
 
 ##### Enhancements
 
-* Add `encoding` option to `Parser` as `Parser.Encoding` type for select
-  encoding on calling CYaml API. Along with that change, `encoding` option
-  are added to `load()`, `load_all()`, `compose()`, `compose_all` and 
-  `YAMLDecoder`. Default value of `encoding` option will be selected by
-  runtime environment.   
+* Add `encoding` option to `Parser` as `Parser.Encoding` type to specify
+  which encoding to pass to libYAML. Along with that change, add `encoding`
+  options to `load()`, `load_all()`, `compose()`, `compose_all()` and 
+  `YAMLDecoder`. The default encoding will be determined at run time based on
+  the String type's native encoding.  
   [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Bug Fixes
--- a/Sources/Yams/Decoder.swift
+++ b/Sources/Yams/Decoder.swift
@@ -12,13 +12,16 @@ import Foundation
 /// user info mapping. Similar to `Foundation.JSONDecoder`.
 public class YAMLDecoder {
     /// Creates a `YAMLDecoder` instance.
-    public init() {}
+    ///
+    /// - parameter encoding: Encoding, `.default` if omitted.
+    public init(encoding: Parser.Encoding = .default) {}
 
     /// Decode a `Decodable` type from a given `String` and optional user info mapping.
     ///
     /// - parameter type:    `Decodable` type to decode.
     /// - parameter yaml:     YAML string to decode.
     /// - parameter userInfo: Additional key/values which can be used when looking up keys to decode.
+    /// - parameter encoding: `Parser.Encoding`
     ///
     /// - returns: Returns the decoded type `T`.
     ///
@@ -27,7 +30,7 @@ public class YAMLDecoder {
                           from yaml: String,
                           userInfo: [CodingUserInfoKey: Any] = [:]) throws -> T where T: Swift.Decodable {
         do {
-            let node = try Yams.compose(yaml: yaml, .basic) ?? ""
+            let node = try Parser(yaml: yaml, resolver: .basic, encoding: encoding).singleRoot() ?? ""
             let decoder = _Decoder(referencing: node, userInfo: userInfo)
             let container = try decoder.singleValueContainer()
             return try container.decode(type)
@@ -39,6 +42,9 @@ public class YAMLDecoder {
                                                     underlyingError: error))
         }
     }
+
+    /// Encoding
+    public var encoding = Parser.Encoding.default
 }
 
 private struct _Decoder: Decoder {
--- a/Sources/Yams/Decoder.swift
+++ b/Sources/Yams/Decoder.swift
@@ -14,14 +14,15 @@ public class YAMLDecoder {
     /// Creates a `YAMLDecoder` instance.
     ///
     /// - parameter encoding: Encoding, `.default` if omitted.
-    public init(encoding: Parser.Encoding = .default) {}
+    public init(encoding: Parser.Encoding = .default) {
+        self.encoding = encoding
+    }
 
     /// Decode a `Decodable` type from a given `String` and optional user info mapping.
     ///
     /// - parameter type:    `Decodable` type to decode.
     /// - parameter yaml:     YAML string to decode.
     /// - parameter userInfo: Additional key/values which can be used when looking up keys to decode.
-    /// - parameter encoding: `Parser.Encoding`
     ///
     /// - returns: Returns the decoded type `T`.
     ///
@@ -44,7 +45,7 @@ public class YAMLDecoder {
     }
 
     /// Encoding
-    public var encoding = Parser.Encoding.default
+    public var encoding: Parser.Encoding
 }
 
 private struct _Decoder: Decoder {
--- a/Sources/Yams/Parser.swift
+++ b/Sources/Yams/Parser.swift
@@ -155,9 +155,12 @@ public final class Parser {
         switch encoding {
         case .utf8:
             yaml_parser_set_encoding(&parser, YAML_UTF8_ENCODING)
+            buffer = .utf8View(yaml.utf8)
+            if try yaml.utf8.withContiguousStorageIfAvailable(startParse(with:)) == nil {
             let utf8Slice = string.utf8CString.dropLast()
-            buffer = .utf8(utf8Slice)
+                buffer = .utf8Slice(utf8Slice)
             try utf8Slice.withUnsafeBytes(startParse(with:))
+            }
         case .utf16:
             // use native endian
             let isLittleEndian = 1 == 1.littleEndian
@@ -209,10 +212,11 @@ public final class Parser {
     private var parser = yaml_parser_t()
 
     private enum Buffer {
-        case utf8(ArraySlice<CChar>)
+        case utf8View(String.UTF8View)
+        case utf8Slice(ArraySlice<CChar>)
         case utf16(Data)
     }
-    private let buffer: Buffer
+    private var buffer: Buffer
 }
 
 // MARK: Implementation Details
@@ -244,6 +248,11 @@ private extension Parser {
 
     func startParse(with buffer: UnsafeRawBufferPointer) throws {
         yaml_parser_set_input_string(&parser, buffer.baseAddress?.assumingMemoryBound(to: UInt8.self), buffer.count)
+        try parse() // Drop YAML_STREAM_START_EVENT
+    }
+
+    func startParse(with buffer: UnsafeBufferPointer<UInt8>) throws {
+        yaml_parser_set_input_string(&parser, buffer.baseAddress, buffer.count)
         try parse() // Drop YAML_STREAM_START_EVENT
     }
 
@@ -401,6 +410,22 @@ private extension Data {
         return try withUnsafeBytes {
             try apply(UnsafeRawBufferPointer(start: $0, count: count))
         }
+    }
+}
+#endif
+
+#if swift(>=4.2)
+#if !compiler(>=5)
+private extension String.UTF8View {
+    func withContiguousStorageIfAvailable<R>(_ body: (UnsafeBufferPointer<Element>) throws -> R) rethrows -> R? {
+        return nil
+    }
+}
+#endif
+#else
+private extension String.UTF8View {
+    func withContiguousStorageIfAvailable<R>(_ body: (UnsafeBufferPointer<Element>) throws -> R) rethrows -> R? {
+        return nil
     }
 }
 #endif
--- a/Sources/Yams/Parser.swift
+++ b/Sources/Yams/Parser.swift
@@ -127,7 +127,9 @@ public final class Parser {
             let key = "YAMS_ENCODING"
             if let yamsEncoding = ProcessInfo.processInfo.environment[key],
                 let encoding = Encoding(rawValue: yamsEncoding.lowercased()) {
-                print("`Parser.Encoding.default` is set to `\(encoding)` by `\(key)` environment variable.")
+                print("""
+                    `Parser.Encoding.default` was set to `\(encoding)` by the `\(key)` environment variable.
+                    """)
                 return encoding
             }
             return "test".utf8.withContiguousStorageIfAvailable({ _ in true }) != nil ? .utf8 : .utf16
--- a/Sources/Yams/Parser.swift
+++ b/Sources/Yams/Parser.swift
@@ -134,7 +134,7 @@ public final class Parser {
                     """)
                 return encoding
             }
-            return "test".utf8.withContiguousStorageIfAvailable({ _ in true }) != nil ? .utf8 : .utf16
+            return key.utf8.withContiguousStorageIfAvailable({ _ in true }) != nil ? .utf8 : .utf16
         }()
     }
     /// Encoding
--- a/Sources/Yams/Parser.swift
+++ b/Sources/Yams/Parser.swift
@@ -126,8 +126,10 @@ public final class Parser {
         public static var `default`: Encoding = {
         #if USE_UTF8
             return .utf8
-        #else
+        #elseif USE_UTF16
             return .utf16
+        #else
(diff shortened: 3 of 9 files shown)


norio-nomura:
> - [ ] ~~Switch Emitter's preferred encoding on runtime~~

I decided to remove `USE_UTF8` compiler condition from `Emitter`, and changed `Emitter` to always use `YAML_UTF8_ENCODING`.

>- [x] Optimize performance on Swift 5.0

I changed`Parser.Encoding.default` to returns `.utf8` If `String.UTF8View.withContiguousStorageIfAvailable(_:)` works.

Sources/Yams/Parser.swift
```diff
@@ -108,33 +116,62 @@ public final class Parser {
...
+        case .utf8:
+            yaml_parser_set_encoding(&parser, YAML_UTF8_ENCODING)
+            buffer = .utf8View(yaml.utf8)
+            if try yaml.utf8.withContiguousStorageIfAvailable(startParse(with:)) == nil {
+                let utf8Slice = string.utf8CString.dropLast()
+                buffer = .utf8Slice(utf8Slice)
+                try utf8Slice.withUnsafeBytes(startParse(with:))
+            }
+        case .utf16:
+            // use native endian
```
jpsim: Nit:
```suggestion
            // use native endianness
```

Sources/Yams/Parser.swift
```diff
@@ -108,33 +116,62 @@ public final class Parser {
...
-        yaml_parser_set_encoding(&parser, isLittleEndian ? YAML_UTF16LE_ENCODING : YAML_UTF16BE_ENCODING)
-        let encoding: String.Encoding = isLittleEndian ? .utf16LittleEndian : .utf16BigEndian
-        data = yaml.data(using: encoding)!
-        try data.withUnsafeBytes(startParse(with:))
-#endif
+        switch encoding {
+        case .utf8:
+            yaml_parser_set_encoding(&parser, YAML_UTF8_ENCODING)
+            buffer = .utf8View(yaml.utf8)
+            if try yaml.utf8.withContiguousStorageIfAvailable(startParse(with:)) == nil {
```
jpsim: It took me a while to understand what was happening here and why we weren't handling the `else` branch.

A comment or code restructure would be useful to indicate that if the result is non-nil, the operation was successful, and the nil case falls back to converting the string to a UTF8 slice.

Suggestion 1:

```swift
let utf8View = yaml.utf8
buffer = .utf8View(utf8View)
if try utf8View.withContiguousStorageIfAvailable(startParse(with:)) != nil {
    // Attempt to parse with underlying UTF8 String encoding was successful, nothing further to do
} else {
    // Fall back to using UTF8 slice
    let utf8Slice = string.utf8CString.dropLast()
    buffer = .utf8Slice(utf8Slice)
    try utf8Slice.withUnsafeBytes(startParse(with:))
}
```

Or suggestion 2:

```swift
let utf8View = yaml.utf8
buffer = .utf8View(utf8View)
let parsingFailed = try utf8View.withContiguousStorageIfAvailable(startParse(with:)) == nil
guard parsingFailed else { break }

// Attempt to parse with underlying UTF8 String encoding failed.
// Fall back to using UTF8 slice.
let utf8Slice = string.utf8CString.dropLast()
buffer = .utf8Slice(utf8Slice)
try utf8Slice.withUnsafeBytes(startParse(with:))
```

Sources/Yams/YamlError.swift
```diff
@@ -81,6 +81,13 @@ public enum YamlError: Swift.Error {
             return text + " in line \(mark.line), column \(mark.column)\n"
         }
     }
+
+#if swift(>=4.1.50)
+    @available(*, unavailable, renamed: "reader(problem:offset:value:yaml:)")
+    public static func reader(problem: String, byteOffset: Int, value: Int32, yaml: String) {
```
jpsim: Should we return `YamlError` here to mimic a case member? And should we proxy this to the new member? Honest questions.

```swift
public static func reader(problem: String, byteOffset: Int, value: Int32, yaml: String) -> YamlError {
    return .reader(problem: problem, offset: byteOffset, value: value, yaml: yaml)
}
```

Sources/Yams/YamlError.swift
```diff
@@ -23,11 +23,11 @@ public enum YamlError: Swift.Error {
     // Used in `yaml_parser_t`
     /// `YAML_READER_ERROR`. Cannot read or decode the input stream.
     ///
-    /// - parameter problem:    Error description.
-    /// - parameter byteOffset: The byte about which the problem occured.
-    /// - parameter value:      The problematic value (-1 is none).
-    /// - parameter yaml:       YAML String which the problem occured while reading.
-    case reader(problem: String, byteOffset: Int, value: Int32, yaml: String)
+    /// - parameter problem: Error description.
+    /// - parameter offset:  The offset about which the problem occured.
```
jpsim: I think it's important to document the unit of the offset here. From my understanding, it's this, but let me know if I'm wrong or if this description can be improved:

> The offset at which the problem occurred, in number of code units in the encoding used by the parser.

Sources/Yams/YamlError.swift
```diff
@@ -142,11 +160,12 @@ extension YamlError: CustomStringConvertible {
             return "No error is produced"
         case .memory:
             return "Memory error"
-        case let .reader(problem, byteOffset, value, yaml):
-            guard let (mark, contents) = markAndSnippet(from: yaml, byteOffset)
-                else { return "\(problem) at byte offset: \(byteOffset), value: \(value)" }
+        case let .reader(problem, offset, value, yaml):
+            guard let (line, column, contents) = offset.flatMap(yaml.lineNumberColumnAndContents(at:))
```
jpsim: Nit: I find this much more readable if the body is indented

```swift
guard let (line, column, contents) = offset.flatMap(yaml.lineNumberColumnAndContents(at:)) else {
    return "\(problem) at offset: \(String(describing: offset)), value: \(value)"
}

let mark = Mark(line: line + 1, column: column + 1)
return "\(mark): error: reader: \(problem):\n" + contents.endingWithNewLine
    + String(repeating: " ", count: column) + "^"
```

Tests/YamsTests/PerformanceTests.swift
```diff
@@ -57,90 +57,137 @@ class PerformanceTests: XCTestCase {
         "/SourceKitten/Source/SourceKittenFramework/Xcode.swift"
     ]
 
-    func testSourceKittenIssue289UsingLoad() {
-        guard let yamlString = try? String(contentsOfFile: filename, encoding: .utf8) else {
+    func loadYAML() throws -> String {
+        let data = try Data(contentsOf: URL(fileURLWithPath: filename))
```
jpsim: Why not use `String(contentsOfFile:encoding:)` like before?

```swift
func loadYAML() throws -> String {
    return try String(contentsOfFile: filename, encoding: .utf8)
}
```

Tests/YamsTests/PerformanceTests.swift
```diff
@@ -57,90 +57,137 @@ class PerformanceTests: XCTestCase {
...
+            }
+            XCTAssertEqual(imports, self.expectedImports)
+            XCTAssertEqual(otherArguments, self.expectedOtherArguments)
+            XCTAssertEqual(sources, self.expectedSources)
+        } catch {
+            XCTFail("\(error)")
+        }
+    }
+
+    func testUsingLoadWithUTF16() {
```
jpsim: XCTest tests can be marked as `throws`, and if a test throws, it's marked as a test failure.

```swift
func testUsingLoadWithUTF16() throws {
    let yaml = try loadYAML()
    self.measure {
        parseSourceKittenIssue289UsingLoad(yaml: yaml, spmName: "SourceKittenFramework", encoding: .utf16)
    }
}
```

Tests/YamsTests/PerformanceTests.swift
```diff
@@ -57,90 +57,137 @@ class PerformanceTests: XCTestCase {
...
-                    let commands = (yaml["commands"] as? [String: [String: Any]])?.values,
-                    let moduleCommand = commands.first(where: { ($0["module-name"] as? String ?? "") == spmName }),
-                    let imports = moduleCommand["import-paths"] as? [String],
-                    let otherArguments = moduleCommand["other-args"] as? [String],
-                    let sources = moduleCommand["sources"] as? [String] else {
+            parseSourceKittenIssue289UsingLoad(yaml: yaml, spmName: "SourceKittenFramework", encoding: .utf16)
+        }
+    }
+
+    func testUsingLoadWithUTF8() {
```
jpsim: Same here about using throwing tests:

```swift
func testUsingLoadWithUTF8() throws {
    let yaml = try loadYAML()
    self.measure {
        parseSourceKittenIssue289UsingLoad(yaml: yaml, spmName: "SourceKittenFramework", encoding: .utf8)
    }
}
```

Tests/YamsTests/PerformanceTests.swift
```diff
@@ -57,90 +57,137 @@ class PerformanceTests: XCTestCase {
...
+        guard let yaml = try? loadYAML() else {
+            XCTFail("Can't load \(filename)")
+            return
+        }
+        self.measure {
+            parseSourceKittenIssue289UsingLoad(yaml: yaml, spmName: "SourceKittenFramework", encoding: .utf8)
+        }
+    }
+
+    func parseSourceKittenIssue289UsingCompose(yaml: String, spmName: String, encoding: Parser.Encoding) {
```
jpsim: All `spmName:` parameters in this file are passed the same value, so you could simplify by just inlining the module name in the functions.

azure-pipelines.yml
```diff
@@ -50,9 +50,13 @@ jobs:
   container: $[ variables['containerResource'] ]
   steps:
     - script: swift test --parallel
-      displayName: swift test
-    - script: swift test --parallel -Xswiftc -DUSE_UTF8
+      displayName: swift test with UTF16
+      env:
+        YAMS_ENCODING: utf16
```
jpsim: Nit: I'd refer to encodings in their usual all-caps capitalization whenever possible, with exceptions when it would violate Swift's API naming guidelines (e.g. case members or identifiers).

```suggestion
        YAMS_ENCODING: UTF16
```

Same with the other environment variables specified in this file.

Sources/Yams/Decoder.swift
```diff
@@ -12,13 +12,16 @@ import Foundation
 /// user info mapping. Similar to `Foundation.JSONDecoder`.
 public class YAMLDecoder {
     /// Creates a `YAMLDecoder` instance.
-    public init() {}
+    ///
+    /// - parameter encoding: Encoding, `.default` if omitted.
+    public init(encoding: Parser.Encoding = .default) {}
```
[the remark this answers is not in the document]
norio-nomura: Oops!

Sources/Yams/Parser.swift
```diff
@@ -108,33 +116,62 @@ public final class Parser {
...
+        /// Use `YAML_UTF8_ENCODING`
+        case utf8
+        /// Use `YAML_UTF16(BE|LE)_ENCODING`
+        case utf16
+        /// Default encoding can be selected at compile time
+        public static var `default`: Encoding = {
+            let key = "YAMS_ENCODING"
+            if let yamsEncoding = ProcessInfo.processInfo.environment[key],
+                let encoding = Encoding(rawValue: yamsEncoding.lowercased()) {
+                print("`Parser.Encoding.default` is set to `\(encoding)` by `\(key)` environment variable.")
```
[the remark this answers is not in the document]
norio-nomura: 🙏

Sources/Yams/Parser.swift
```diff
@@ -108,33 +116,62 @@ public final class Parser {
     /// Constructor.
     public let constructor: Constructor
 
+    /// Encoding
+    public enum Encoding: String {
+        /// Use `YAML_UTF8_ENCODING`
+        case utf8
+        /// Use `YAML_UTF16(BE|LE)_ENCODING`
+        case utf16
+        /// Default encoding can be selected at compile time
```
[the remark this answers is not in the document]
norio-nomura: Thanks! 🙏

Sources/Yams/Parser.swift
norio-nomura: 🙏

Sources/Yams/Parser.swift
norio-nomura: Thanks for suggestions! 🙏  I prefer the former.

Sources/Yams/YamlError.swift
norio-nomura: > Should we return `YamlError` here to mimic a case member?

Calculating `offset` from `byteOffset` requires parser's encoding, but it is unknown at here.
So we can't return it.

Before this PR, parser's encoding was determined at compile time, so `byteOffset` could be converted to `line` and `column` in yaml string on generating `YamlError.description` later.
This PR changes `YamlError.init(from:with:)` to converting byte offset into `String.IndexDistance?` with using `parser`'s encoding.

Sources/Yams/YamlError.swift
norio-nomura: > in number of code units in the encoding used by the parser.

It is `String.IndexDistance?` since converting byte offset to `String.Index` may fail.
So, it may be not in number of code units in the encoding used by the parser.
🤔 

Currently, `Parser` receives `String` so it should not fail here. But in the future `Parser` may receive `Data`, in which case it may actually fail.

Tests/YamsTests/PerformanceTests.swift
norio-nomura: As far as I tested, `String.UTF8View.withContiguousStorageIfAvailable(_:)` does not work if `String` is created by `String.init(contentsOfFile:)` on Apple Platform.

jpsim: That seems odd but okay.

Sources/Yams/YamlError.swift
jpsim: 👍 thanks for explaining!

norio-nomura:
Even if I push a new commit to PR, it seems that Azure Pipelines does not start CI automatically.

norio-nomura:
I manually triggered Azure Pipelines from Checks tab.

Sources/Yams/Parser.swift
```diff
@@ -214,6 +261,11 @@ private extension Parser {
         try parse() // Drop YAML_STREAM_START_EVENT
     }
 
+    func startParse(with buffer: UnsafeBufferPointer<UInt8>) throws {
```
jpsim:
```suggestion
    private func startParse(with buffer: UnsafeBufferPointer<UInt8>) throws {
```

Same with the other `startParse(...)` function.

Sources/Yams/Parser.swift
```diff
@@ -108,33 +116,70 @@ public final class Parser {
...
+        public static var `default`: Encoding = {
+            let key = "YAMS_DEFAULT_ENCODING"
+            if let yamsEncoding = ProcessInfo.processInfo.environment[key],
+                let encoding = Encoding(rawValue: yamsEncoding.lowercased()) {
+                print("""
+                    `Parser.Encoding.default` was set to `\(encoding)` by the `\(key)` environment variable.
+                    """)
+                return encoding
+            }
+            return "test".utf8.withContiguousStorageIfAvailable({ _ in true }) != nil ? .utf8 : .utf16
```
jpsim: I don't feel strongly, but we could use `key.utf8...` here instead to avoid adding another string to the binary.

Sources/Yams/String+Yams.swift
```diff
@@ -11,31 +11,15 @@ import Foundation
...
-    ///
-    /// - returns: lineNumber: line number start from 0,
-    ///            column: utf16 column start from 0,
-    ///            contents: substring of line
-    func utf16LineNumberColumnAndContents(at offset: Int) -> LineNumberColumnAndContents? {
-        guard let index = utf16
-            .index(utf16.startIndex, offsetBy: offset, limitedBy: utf16.endIndex)?
-            .samePosition(in: self) else { return nil }
+    func lineNumberColumnAndContents(at offset: Int) -> LineNumberColumnAndContents? {
+        guard let index = index(startIndex, offsetBy: offset, limitedBy: endIndex) else { return nil }
```
jpsim: Nit:

```swift
func lineNumberColumnAndContents(at offset: Int) -> LineNumberColumnAndContents? {
    return index(startIndex, offsetBy: offset, limitedBy: endIndex).flatMap(lineNumberColumnAndContents)
}
```

norio-nomura:
> I think this is because of the brownout: https://www.githubstatus.com/incidents/kq75qp7try8t

Azure Pipelines uses GitHub Apps and does not use GitHub Services. Hm… 🤔

Sources/Yams/Parser.swift
norio-nomura: Those `private` modifier may be redundant since they are declared inside private extension of `Parser`. Swift 5.0 warns `public` modifier are redundant in public extension, but not on `private` modifier in private extension.

jpsim: Oh all good then!

Sources/Yams/YamlError.swift
norio-nomura: How do you think about following?
```diff
-    /// - parameter offset:  The offset about which the problem occured.
+    /// - parameter offset:  The offset from `yaml.startIndex` about which the problem occured.
     /// - parameter value:   The problematic value (-1 is none).
     /// - parameter yaml:    YAML String which the problem occured while reading.
-    case reader(problem: String, offset: Int?, value: Int32, yaml: String)
+    case reader(problem: String, offset: String.IndexDistance?, value: Int32, yaml: String)
```

jpsim: Great! Other than changing "about which" to "at which".

CHANGELOG.md
```diff
@@ -2,11 +2,17 @@
...
 ##### Breaking
 
-* None.
+* Change `byteOffset` to `offset` in `YamlError.reader`.  
+  [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Enhancements
 
-* None.
+* Add `encoding` option to `Parser` as `Parser.Encoding` type for select
```
jpsim: Suggested changes:

```markdown
* Add `encoding` option to `Parser` as `Parser.Encoding` type to specify
  which encoding to pass to libYAML. Along with that change, add `encoding`
  options to `load()`, `load_all()`, `compose()`, `compose_all()` and 
  `YAMLDecoder`. The default encoding will be determined at run time based on
  the String type's native encoding.  
  [Norio Nomura](https://github.com/norio-nomura)
```

norio-nomura: Thanks! 🙏

Sources/Yams/YamlError.swift
norio-nomura: I forgot that  'IndexDistance' is deprecated.

norio-nomura:
Thanks! 🙏

PR #146, opened 2019-01-05
[Swift 5.0] Update issues related `nanosecond` in `Date` from swift-corelibs-foundation

norio-nomura:
- Change `os(Linux)` compilation condition to `_runtime(_ObjC)`
- [SR-3158](https://bugs.swift.org/browse/SR-3158) has been fixed on Swift 5.0
- [SR-6223](https://bugs.swift.org/browse/SR-6223) has been fixed on Swift 5.0
- Remove two `FIXME`

PR #147, opened 2019-01-06
Add unused imports CI job

jpsim:


jpsim:
Making sure CI fails and that removing unused imports will make it pass.

jpsim:
Unrelated, I never noticed how much faster UTF8 was than UTF16

image

jpsim:
Also, it looks like Azure Pipelines is running automatically whenever I create or push commits to a PR. Maybe it's a permission issue that you need to manually trigger @norio-nomura ?

norio-nomura:
> Maybe it's a permission issue that you need to manually trigger @norio-nomura ?

Sorry, I forgot to report. On working #146, I only needed the trigger manually once, CI was automatically started afterwards. 🙇

norio-nomura:
> Unrelated, I never noticed how much faster UTF8 was than UTF16

The difference in processing time comes from whether it is compiled or not. `swift test with UTF8` job does not compile at all and only performs tests.
The actual speed difference due to encoding is not so much observable from the processing time of the whole test.

PR #153, opened 2019-01-18
Update Azure Pipelines configuration

norio-nomura:


jpsim:
👍 when you get the analyze job to pass

PR #154, opened 2019-01-21
[WIP] Add Windows CI job

jpsim:
There's no way this will work, but worth a shot.

Getting the Swift toolchain from https://forums.swift.org/t/windows-nightlies/19174


norio-nomura:
If we imitate [an example](https://docs.microsoft.com/en-us/azure/devops/pipelines/scripts/cross-platform-scripting?view=vsts&tabs=yaml#switch-based-on-platform) of setting the job scope environment variable, it would be as follows:
```yaml
    - powershell: Write-Host "##vso[task.setvariable variable=PATH]$PATH;C:\Swift\toolchain\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin"
      displayName: Add Swift to PATH
```
SeeAlso: [Set a job-scoped variable from a script](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=vsts&tabs=yaml%2Cbatch#set-a-job-scoped-variable-from-a-script)

norio-nomura:
Sorry, `\` needs to be escaped in double quote. :(
https://github.com/compnerd/swift-windows-build/blob/master/azure-pipelines.yml#L49-L51
We should have checked that link were mentioned on https://forums.swift.org/t/windows-nightlies/19174.

azure-pipelines.yml
```diff
@@ -2,185 +2,19 @@ trigger:
...
-      displayName: Validate documentation coverage
-    - task: PublishPipelineArtifact@0
-      inputs:
-        artifactName: 'API Docs'
-        targetPath: 'docs'
+    - powershell: Invoke-WebRequest 'https://compnerd.visualstudio.com/3133d6ab-80a8-4996-ac4f-03df25cd3224/_apis/build/builds/109/artifacts?artifactName=toolchain&api-version=5.0&%24format=zip' -outfile .\toolchain.zip
+      displayName: Download Swift toolchain for Windows
+    - powershell: Expand-Archive toolchain.zip -DestinationPath C:\Swift
+      displayName: Install Swift
+    - powershell: Write-Host "##vso[task.setvariable variable=PATH]$PATH;C:\\Swift\\toolchain\\Developer\\Toolchains\\unknown-Asserts-development.xctoolchain\\usr\\bin"
```
norio-nomura: powershell does not seem to recognize `$PATH`.
```suggestion
    - powershell: Write-Host "##vso[task.setvariable variable=PATH]%PATH%;C:\\Swift\\toolchain\\Developer\\Toolchains\\unknown-Asserts-development.xctoolchain\\usr\\bin"
```

jpsim:
> it did not contain `swift-build` or `swift-test`

Well then this effort is pointless. I didn’t bother virtualizing Windows to try, I was going to see when I got to that point in CI to see if it worked.

Thanks for your help!

jpsim:
We can try this again when SwiftPM works on Windows.

PR #158, opened 2019-01-29
_KeyedDecodingContainer and _UnkeyedDecodingContainer did not properly decode null

### LINKED ISSUE #157: null/~/NULL/Null parsed as strings, not nil
I'm probably just holding it wrong but when I try to decode yml values with nil values - which I believe are represented as `null | ~ | NULL | Null` in yml - I get the corresponding `String`s back and not nil:

```
    func test_null_yml() throws {
        let s = """
              n1: ~
              n2: null
              n3: NULL
              n4: Null
              n5:
            """
        struct Test: Decodable {
            let n1: String?
            let n2: String?
            let n3: String?
            let n4: String?
            let n5: String?
        }
        let t = try YAMLDecoder().decode(Test.self, from: s)
        XCTAssertNil(t.n1)  // XCTAssertNil failed: "~"
        XCTAssertNil(t.n2)  // XCTAssertNil failed: "null"
        XCTAssertNil(t.n3)  // XCTAssertNil failed: "NULL"
        XCTAssertNil(t.n4)  // XCTAssertNil failed: "Null"
        XCTAssertNil(t.n5)  // XCTAssertNil failed: ""
    }
```

This is with swift-4.2.1 and Yams

```
          "revision": "26ab35f50ea891e8edefcc9d975db2f6b67e1d68",
          "version": "1.0.1"
```

I see references to all these `null` tags in 
https://github.com/jpsim/Yams/blob/c64a5f2c68a9bd13272e3f665c969144a8d67eb4/Tests/YamsTests/ResolverTests.swift#L16 and so I'm wondering: is there something wrong with my test setup above? Or is this perhaps a 🐛?

norio-nomura:
Fixes #157

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@
   [Norio Nomura](https://github.com/norio-nomura)
   [#146](https://github.com/jpsim/Yams/pull/146)
 
+* Fix null/~/NULL/Null were parsed as strings, not nil by `YAMLDecoder`.  
+  [Norio Nomura](https://github.com/norio-nomura)
+  [#157](https://github.com/jpsim/Yams/issues/157)
+
 ## 1.0.1
 
 ##### Breaking
--- a/Sources/Yams/Decoder.swift
+++ b/Sources/Yams/Decoder.swift
@@ -116,7 +116,7 @@ private struct _KeyedDecodingContainer<Key: CodingKey> : KeyedDecodingContainerP
     func contains(_ key: Key) -> Bool { return mapping[key.stringValue] != nil }
 
     func decodeNil(forKey key: Key) throws -> Bool {
-        return try node(for: key) == Node("null", Tag(.null))
+        return try decoder(for: key).decodeNil()
     }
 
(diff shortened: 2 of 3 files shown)


norio-nomura:
Yes, we can.
Which commit should we branch from? c935ba12b180da1d3a2ecc07bb8ba7e63627eddc?

jpsim:
Sure, sounds good. We'll need to pull in the CI changes too if we want to run that branch on CI.

finestructure:
Thanks a lot for merging in that fix! I don't think there's a release with that, is there?

I'd rather not pull in `master` instead of tags so I'm wondering if I should tag in a fork. Or do you plan on pushing out a release soon?

jpsim:
I'll make a 1.0.2 release today with the tip of `yams-1.0-branch`. `master` has some breaking changes so we'll want to cut a 2.0 release from that shortly too.

jpsim:
1.0.2 is out https://github.com/jpsim/Yams/releases/tag/1.0.2

Issue #174, opened 2019-02-11
Yams.serialize converts string in ' to number


sbarow:
Following on #116 when trying to serialise something like`value: '10.0'` the output will be `value: 10.0`.

Unfortunately doing `value: !!str '10.0'` still results in the output being  number,  tested with `"` and the same happens.

Example updating:

```
let updatedNode = try node.represented()
let serialized = try Yams.serialize(node: updatedNode)
try path.write(serialized)
```

sbarow:
Doesn't look like this is specific to `Yams` but rather `CYaml` 

`Emitter.swift`
```
    private func serializeScalar(_ scalar: Node.Scalar) throws {
	...
        let scalar_style = yaml_scalar_style_t(rawValue: scalar.style.rawValue) // Returns `any` as that's what's passed in.
	...
    }
```

Looking at this, a possible solution could be in `Representer.swift` with something like this:

```
private func represent(_ value: Any, _ isValue: Bool = true) throws -> Node {
    if let string = value as? String {
        if value is String && isValue {
            return Node(string, .implicit, .singleQuoted)
        }
        return Node(string)
    } else if let representable = value as? NodeRepresentable {
        return try representable.represented()
    }
    throw YamlError.representer(problem: "Failed to represent \(value)")
}
```

If you're okay with this I can put up a PR.

norio-nomura:
Thank you for filing an issue!

Could you provide test code to reproduce the issue?
The code I tried is as follows:
```swift
    func testSrializeNumberInSingleQuote() throws {
        let yaml = """
            value1: '10.0'
            value2: !!str '10.0'

            """
        let node = try Yams.compose(yaml: yaml)
        let updatedNode = try node.represented()
        let serialized = try Yams.serialize(node: updatedNode)
        print(serialized)
    }
```
result:
```console
Test Suite 'Selected tests' started at 2019-02-16 09:23:46.575
Test Suite 'YamsTests.xctest' started at 2019-02-16 09:23:46.576
Test Suite 'EmitterTests' started at 2019-02-16 09:23:46.578
Test Case '-[YamsTests.EmitterTests testSrializeNumberInSingleQuote]' started.
value1: '10.0'
value2: '10.0'

Test Case '-[YamsTests.EmitterTests testSrializeNumberInSingleQuote]' passed (0.135 seconds).
Test Suite 'EmitterTests' passed at 2019-02-16 09:23:46.714.
[...]

PR #175, opened 2019-02-11
Fix conversion of string to float

sbarow:
I don't think this fixes the case when you do `Node("4.2")`, this will still be turned into `4.2` instead of `"4.2"`.


jpsim:
Thanks for the PR @sbarow, although I don't love bypassing what libYAML is doing like this. Can you share steps to reproduce the problem you were seeing in #174 so we can try to investigate the root cause here?

norio-nomura:
#198 is replacing this.

PR #182, opened 2019-04-11
Run on Xcode 10.2 on CI

jpsim:
This was just added to Azure Pipelines: https://github.com/Microsoft/azure-pipelines-image-generation/issues/815#issuecomment-482304766

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,6 +24,8 @@ jobs:
         containerImage: norionomura/swiftlint:swift-4.2.0
       swift421:
         containerImage: norionomura/swiftlint:swift-4.2.1
+      swift50:
+        containerImage: swift:5.0
   container: $[ variables['containerImage'] ]
   steps:
     - script: swift test --parallel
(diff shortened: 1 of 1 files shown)


jpsim:
Need to fix the hanging performance test on Xcode 10.2.

azure-pipelines.yml
```diff
@@ -74,12 +63,85 @@ jobs:
...
+        YAMS_DEFAULT_ENCODING: UTF8
+    - task: PublishTestResults@2
+      inputs:
+        testRunner: JUnit
+        testResultsFiles: build/reports/**
+      condition: succeededOrFailed()
+
+- job: Xcode10
+  pool:
+    vmImage: 'macOS 10.14'
```
norio-nomura: `vmImage` can be made a parameter like https://github.com/norio-nomura/Base32/pull/47/commits/20fcc9b2957f63e54d3ed3cdc05fbb01513ce4a8

jpsim: 😻

norio-nomura:
> Need to fix the hanging performance test on Xcode 10.2.

This can not be reproduced locally. 🤔 
Can you try rebuild?

jpsim:
> This can not be reproduced locally. 🤔
> Can you try rebuild?

Happened again on rebuild. I think the problem is that there may not be any 32 bit simulators installed, or maybe they're not installed properly, and we're only seeing it hang for performance tests because the other ones don't require booting the simulator because they're just logic tests?

PR #188, opened 2019-06-21
[Azure Pipelines] Add scheduled job

norio-nomura:
https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,8 +1,16 @@
 trigger:
 - master
 
+schedules:
+- cron: "0 4 * * *"
+  displayName: 
+  branches:
+    include: 
+    - master
+
 jobs:
 - job: linux
+  condition: ne(variables['Build.Reason'], 'Schedule')
   pool:
     vmImage: 'Ubuntu 16.04'
   strategy:
@@ -38,6 +46,7 @@ jobs:
         testResultsFiles: build/reports/**
 
 - job: macOS
+  condition: ne(variables['Build.Reason'], 'Schedule')
   pool:
     vmImage: $[ coalesce(variables['VM_IMAGE'], 'macOS 10.13') ]
   strategy:
@@ -125,6 +134,7 @@ jobs:
       condition: succeededOrFailed()
 
 - job: CocoaPods
+  condition: ne(variables['Build.Reason'], 'Schedule')
   pool:
     vmImage: 'macOS 10.14'
   variables:
@@ -149,6 +159,7 @@ jobs:
       displayName: Run SwiftLint Analyze
 
 - job: jazzy
+  condition: ne(variables['Build.Reason'], 'Schedule')
   pool:
     vmImage: 'Ubuntu 16.04'
   container: norionomura/jazzy:0.9.6_swift-4.2.4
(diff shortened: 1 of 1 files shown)


norio-nomura:
How do we know that the Nightly build has failed? 🤔

jpsim:
Are these jobs passing right now?

> How do we know that the Nightly build has failed? 🤔

I think the author of the commit will get emailed.

norio-nomura:
> Are these jobs passing right now?

Yes, on my local.
Hm, it may be better to make Nightly also work with ordinary PR.

jpsim:
> Hm, it may be better to make Nightly also work with ordinary PR.

Why? That would encourage merging things that fail CI. I'd rather only merge green PRs, you?

norio-nomura:
Even testing `norionomura/swift:nightly` seems overkill, so I'll commit changes `linux swift51` as scheduled job.

norio-nomura:
Now, `linux swift51` job will triggered by both of PR and schedule.

norio-nomura:
All jobs except `linux swift 51` are triggered by PR.

norio-nomura:
> I think the author of the commit will get emailed.

Does this mean that I will not be notified if the commit you merged is the latest on the master?

jpsim:
Yes, that's how I think it works. As long as one of us has the latest commit on master I think that should be fine.

norio-nomura:
Ok, let's think again if the problem actually occurs. 👍

norio-nomura:
I will check if the job actually started after one and a half hours.

norio-nomura:
Even with the change of #189, the scheduled job did not work. 😞 
https://dev.azure.com/jpsim/Yams/_build?definitionId=2

I found following in [Q & A](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#my-build-didnt-run-what-happened):
> ### My build didn't run. What happened?
>
>Someone must view a page in your organization regularly for CI and scheduled builds to run. It can be any page, including, for example, **Azure Pipelines**.
>
>Your organization goes dormant five minutes after the last user signed out of Azure DevOps. After that, each of your build pipelines will run one more time. For example, while your organization is dormant:
>
> * A nightly build of code in your organization will run only one night until someone signs in again.
>
> * CI builds of an Other Git repo will stop running until someone signs in again.

Can this explain why schedule jobs do not work? 🤔

jpsim:
I suppose that could explain it. If we really wanted to we could set up a bot that runs nightly tests. Not worth the effort IMO.

norio-nomura:
According to the Developer Community post, editing yaml from the Azure GUI will make the schedule work. 😯
https://developercommunity.visualstudio.com/comments/623549/view.html

norio-nomura:
confirmed that the scheduled job started to work.
https://dev.azure.com/jpsim/Yams/_build/results?buildId=785

jpsim:
Lovely! When the 5.1 branch starts working again, we’ll know when it breaks more quickly.

PR #189, opened 2019-06-21
[Azure Pipelines] Add `always: true` to `schedules`

norio-nomura:
>  always: boolean # whether to always run the pipeline or only if there have been source code changes since the last run. The default is false.


jpsim:
👍 makes sense

PR #193, opened 2019-06-25
Enhance handling of `Date` with `nanosecond`

### LINKED ISSUE #192: `YAMLEncoder` outputs a result that is delayed by 1 second when encoding `Date` with nanoseconds greater than 999499997
`DateFormatter` doesn't support time more precise than milliseconds, and smaller digits will be rounded to milliseconds.
Since `DateComponents` having a value of 999499997 or greater in `nanosecond` generates `Date` in which `nanosecond` is 999500036 or greater, `DateFormatter` rounds up and advances the result by 1 second.

POC:
```swift
import Foundation

let iso8601WithFractionalSecondFormatter: DateFormatter = {
    let formatter = DateFormatter()
    formatter.locale = Locale(identifier: "en_US_POSIX")
    formatter.dateFormat = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.SSS'Z"
    formatter.timeZone = TimeZone(secondsFromGMT: 0)
    return formatter
}()

let gregorian = Calendar(identifier: .gregorian)
let utc = TimeZone(identifier: "UTC")!
let components = gregorian.dateComponents(in: utc, from: Date())

func check(from components: DateComponents, with nanosecond: Int) {
    print("--")
    print("Dircted nanosecond: \(nanosecond)")
    var components = components
    components.nanosecond = nanosecond
    let date = components.date!
    print("Actual nanosecond : \(gregorian.dateComponents(in: utc, from: date).nanosecond!)")
    print("Date: \(iso8601WithFractionalSecondFormatter.string(from: date))")
}

check(from: components, with: 999499976)
check(from: components, with: 999499977)
```
Result:
```log
--
Dircted nanosecond: 999499976
Actual nanosecond : 999499917
Date: 2019-06-25T14:35:36.999Z
--
Dircted nanosecond: 999499977
Actual nanosecond : 999500036
Date: 2019-06-25T14:35:37.000Z
```
We should drop `nanosecond` from `Date` before passing to `DateFormatter` when using nanosecond precision.

norio-nomura:
- Support `Date` with `nanosecond` in Swift 4.x.  
- Fix encoding `Date` with `nanosecond` greater than 999499977 (#192)
- Performance up on handling `Date`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@
 
 ##### Bug Fixes
 
-* Fix `YAMLEncoder` outputted a result that is delayed by 1 second when encoding
-  `Date` with `nanosecond` greater than 999499997.  
+* Fix a bug where `YAMLEncoder` would delay `Date`s by 1 second when encoding
+  values with a `nanosecond` component greater than 999499997.    
   [Norio Nomura](https://github.com/norio-nomura)
   [#192](https://github.com/jpsim/Yams/issues/192)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,15 @@
 
 ##### Enhancements
 
-* None.
+* Support `Date` with `nanosecond` in Swift 4.x.  
+  [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Bug Fixes
 
-* None.
+* Fix `YAMLEncoder` outputted a result that is delayed by 1 second when encoding
+  `Date` with `nanosecond` greater than 999499997.  
+  [Norio Nomura](https://github.com/norio-nomura)
+  [#192](https://github.com/jpsim/Yams/issues/192)
 
 ## 2.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ##### Enhancements
 
-* Support `Date` with `nanosecond` in Swift 4.x.  
+* Accurately represent `Date`s with nanosecond components in Swift 4.x.  
   [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Bug Fixes
--- a/Sources/Yams/Constructor.swift
+++ b/Sources/Yams/Constructor.swift
@@ -174,7 +174,7 @@ extension Date: ScalarConstructible {
         datecomponents.hour = components[3].flatMap { Int($0) }
         datecomponents.minute = components[4].flatMap { Int($0) }
         datecomponents.second = components[5].flatMap { Int($0) }
-        datecomponents.nanosecond = components[6].flatMap { fraction in
+        let optionalNanosecond: Int? = components[6].flatMap { fraction in
             let length = fraction.count
             let nanosecond: Int?
             if length < 9 {
@@ -199,7 +199,10 @@ extension Date: ScalarConstructible {
             }
             return TimeZone(secondsFromGMT: seconds)
         }()
-        return datecomponents.date
+        guard let date = datecomponents.date else { return nil }
+        guard let nanosecond = optionalNanosecond else { return date }
+        let timeInterval = date.timeIntervalSinceReferenceDate + TimeInterval(nanosecond) / 1_000_000_000.0
+        return Date(timeIntervalSinceReferenceDate: timeInterval)
     }
 
     private static let timestampPattern: NSRegularExpression = pattern([
--- a/Sources/Yams/Constructor.swift
+++ b/Sources/Yams/Constructor.swift
@@ -167,7 +167,7 @@ extension Date: ScalarConstructible {
         }
 
         var datecomponents = DateComponents()
-        datecomponents.calendar = Calendar(identifier: .gregorian)
+        datecomponents.calendar = gregorianCalendar
         datecomponents.year = components[0].flatMap { Int($0) }
         datecomponents.month = components[1].flatMap { Int($0) }
         datecomponents.day = components[2].flatMap { Int($0) }
@@ -204,6 +204,8 @@ extension Date: ScalarConstructible {
         let timeInterval = date.timeIntervalSinceReferenceDate + TimeInterval(nanosecond) / 1_000_000_000.0
         return Date(timeIntervalSinceReferenceDate: timeInterval)
     }
+
+    private static let gregorianCalendar = Calendar(identifier: .gregorian)
 
     private static let timestampPattern: NSRegularExpression = pattern([
         "^([0-9][0-9][0-9][0-9])",          // year
--- a/Sources/Yams/Constructor.swift
+++ b/Sources/Yams/Constructor.swift
@@ -174,17 +174,17 @@ extension Date: ScalarConstructible {
         datecomponents.hour = components[3].flatMap { Int($0) }
         datecomponents.minute = components[4].flatMap { Int($0) }
         datecomponents.second = components[5].flatMap { Int($0) }
-        let optionalNanosecond: Int? = components[6].flatMap { fraction in
+        let nanoseconds: TimeInterval? = components[6].flatMap { fraction in
             let length = fraction.count
-            let nanosecond: Int?
+            let nanoseconds: Int?
             if length < 9 {
-                nanosecond = Int(fraction).map { number in
+                nanoseconds = Int(fraction).map { number in
                     repeatElement(10, count: 9 - length).reduce(number, *)
                 }
             } else {
-                nanosecond = Int(fraction[..<fraction.index(fraction.startIndex, offsetBy: 9)])
+                nanoseconds = Int(fraction.prefix(9))
             }
-            return nanosecond
+            return nanoseconds.map { Double($0) / 1_000_000_000.0 }
         }
         datecomponents.timeZone = {
             var seconds = 0
@@ -199,10 +199,7 @@ extension Date: ScalarConstructible {
             }
             return TimeZone(secondsFromGMT: seconds)
         }()
-        guard let date = datecomponents.date else { return nil }
-        guard let nanosecond = optionalNanosecond else { return date }
-        let timeInterval = date.timeIntervalSinceReferenceDate + TimeInterval(nanosecond) / 1_000_000_000.0
-        return Date(timeIntervalSinceReferenceDate: timeInterval)
+        return datecomponents.date.map { nanoseconds.map($0.addingTimeInterval) ?? $0 }
     }
 
     private static let gregorianCalendar = Calendar(identifier: .gregorian)
--- a/Sources/Yams/Representer.swift
+++ b/Sources/Yams/Representer.swift
@@ -119,11 +119,18 @@ extension Date: ScalarRepresentable {
         let calendar = Calendar(identifier: .gregorian)
         let nanosecond = calendar.component(.nanosecond, from: self)
         if nanosecond != 0 {
-            return iso8601WithoutZFormatter.string(from: self) +
+            return iso8601WithoutZFormatter.string(from: dateDroppingNanosecond) +
                 String(format: ".%09d", nanosecond).trimmingCharacters(in: characterSetZero) + "Z"
         } else {
             return iso8601Formatter.string(from: self)
         }
+    }
+
+    private var dateDroppingNanosecond: Date {
+        let calendar = Calendar(identifier: .gregorian)
+        var dateComponents = calendar.dateComponents(in: calendar.timeZone, from: self)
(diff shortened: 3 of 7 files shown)


norio-nomura:
will make additional refactoring related to date formatting tomorrow. 😪

Tests/YamsTests/EncoderTests.swift
```diff
@@ -322,6 +322,19 @@ class EncoderTests: XCTestCase { // swiftlint:disable:this type_body_length
         XCTAssertNil(t.n5)
     }
 
+    func testEncodingDateWithNanosecondGreaterThan999499977() throws {
+    #if _runtime(_ObjC) || swift(>=5.0)
+        let gregorian = Calendar(identifier: .gregorian)
+        let utc = TimeZone(identifier: "UTC")!
+        var dateComponents = gregorian.dateComponents(in: utc, from: Date())
```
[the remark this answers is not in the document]
norio-nomura: According to the condition, #192 occurs only 5 times in 10000 times. Nevertheless, I was able to find #192 because I saw that `testEncodingDate()` using `Date()` actually failed. This issue only involves nanoseconds. So I want to use random data as well as `testEncodingDate()` except `nanosecond`.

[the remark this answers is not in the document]
jpsim: Nice catch!

norio-nomura:
> Wouldn’t it be safer and more performant to grab the time interval out of the date, truncate its nanoseconds and then create a new date with that truncated time interval?

I would like to try changing nanosecond handling to `TimeInterval` instead of `DateComponents` for bug evasion and performance.

norio-nomura:
> - Performance up on handling `Date`

In the following test code, `YAMLEncoder.encode()` is about 3 times faster and `YAMLDecoder.decode()` is about 1.3 times faster.
```swift
func testDateRoundtrip() throws {
    var timeInterval = 0.0
    _ = modf(Date().timeIntervalSinceReferenceDate, &timeInterval)
    var nanosecond = 0.999_499_977
    for _ in 0..<100000 {
        nanosecond += 0.000_000_001
        _testRoundTrip(of: Date(timeIntervalSinceReferenceDate: timeInterval + nanosecond))
    }
}
```
before:
スクリーンショット 2019-06-26 19 56 18
after:
スクリーンショット 2019-06-26 19 57 38

Sources/Yams/Constructor.swift
```diff
@@ -199,9 +199,14 @@ extension Date: ScalarConstructible {
             }
             return TimeZone(secondsFromGMT: seconds)
         }()
-        return datecomponents.date
+        guard let date = datecomponents.date else { return nil }
+        guard let nanosecond = optionalNanosecond else { return date }
+        let timeInterval = date.timeIntervalSinceReferenceDate + TimeInterval(nanosecond) / 1_000_000_000.0
```
jpsim: I don't _really_ like `TimeInterval(nanosecond)` here because that doesn't give us an actual time interval.

What do you think of this?

```diff
diff --git a/Sources/Yams/Constructor.swift b/Sources/Yams/Constructor.swift
index e7f7495..0f4863b 100644
--- a/Sources/Yams/Constructor.swift
+++ b/Sources/Yams/Constructor.swift
@@ -174,17 +174,17 @@ extension Date: ScalarConstructible {
         datecomponents.hour = components[3].flatMap { Int($0) }
         datecomponents.minute = components[4].flatMap { Int($0) }
         datecomponents.second = components[5].flatMap { Int($0) }
-        let optionalNanosecond: Int? = components[6].flatMap { fraction in
+        let nanoseconds: TimeInterval? = components[6].flatMap { fraction in
             let length = fraction.count
-            let nanosecond: Int?
+            let nanoseconds: Int?
             if length < 9 {
-                nanosecond = Int(fraction).map { number in
+                nanoseconds = Int(fraction).map { number in
                     repeatElement(10, count: 9 - length).reduce(number, *)
                 }
             } else {
-                nanosecond = Int(fraction[..<fraction.index(fraction.startIndex, offsetBy: 9)])
+                nanoseconds = Int(fraction[..<fraction.index(fraction.startIndex, offsetBy: 9)])
             }
-            return nanosecond
+            return nanoseconds.map { Double($0) / 1_000_000_000.0 }
         }
         datecomponents.timeZone = {
             var seconds = 0
@@ -199,10 +199,7 @@ extension Date: ScalarConstructible {
             }
             return TimeZone(secondsFromGMT: seconds)
         }()
-        guard let date = datecomponents.date else { return nil }
-        guard let nanosecond = optionalNanosecond else { return date }
-        let timeInterval = date.timeIntervalSinceReferenceDate + TimeInterval(nanosecond) / 1_000_000_000.0
-        return Date(timeIntervalSinceReferenceDate: timeInterval)
+        return datecomponents.date.map { $0.addingTimeInterval(nanoseconds ?? 0) }
     }
```

CHANGELOG.md
```diff
@@ -7,11 +7,15 @@
 
 ##### Enhancements
 
-* None.
+* Support `Date` with `nanosecond` in Swift 4.x.
```
jpsim: Suggestion:

```suggestion
* Accurately represent `Date`s with nanosecond components in Swift 4.x.  
```

CHANGELOG.md
```diff
@@ -7,11 +7,15 @@
...
 ##### Enhancements
 
-* None.
+* Support `Date` with `nanosecond` in Swift 4.x.  
+  [Norio Nomura](https://github.com/norio-nomura)
 
 ##### Bug Fixes
 
-* None.
+* Fix `YAMLEncoder` outputted a result that is delayed by 1 second when encoding
```
jpsim: Suggestion:

```md
* Fix a bug where `YAMLEncoder` would delay `Date`s by 1 second when encoding values with a `nanosecond` component greater than 999499997.  
```

Tests/YamsTests/EncoderTests.swift
```diff
@@ -322,6 +308,13 @@ class EncoderTests: XCTestCase { // swiftlint:disable:this type_body_length
         XCTAssertNil(t.n5)
     }
 
+    func testEncodingDateWithNanosecondGreaterThan999499977() throws {
+        var timeInterval = 0.0
+        _ = modf(Date().timeIntervalSinceReferenceDate, &timeInterval)
```
jpsim: Nice

Tests/YamsTests/TestHelper.swift
```diff
@@ -20,17 +19,13 @@ func timestamp(_ timeZoneHour: Int = 0,
...
-                          hour: hour, minute: minute, second: second, nanosecond: nanosecond)
-    // Using `DateComponents.date` causes crash on Linux
-    guard let date = NSCalendar(identifier: .gregorian)?.date(from: datecomponents) else {
-        fatalError("Never happen this")
-    }
-    return date
+    let dateComponents = DateComponents(calendar: gregorianCalendar, timeZone: timeZone,
+                                        year: year, month: month, day: day,
+                                        hour: hour, minute: minute, second: second)
+    guard let date = dateComponents.date else { fatalError("Never happen this") }
```
jpsim: Nit:

```suggestion
    guard let date = dateComponents.date else { fatalError("Tests shouldn't create timestamps for invalid dates") }
```

Tests/YamsTests/TestHelper.swift
```diff
@@ -20,17 +19,13 @@ func timestamp(_ timeZoneHour: Int = 0,
...
+    guard let fraction = fraction else { return date }
```
jpsim: Nit:

```swift
return date.addingTimeInterval(fraction ?? 0)
```

Sources/Yams/Representer.swift
```diff
@@ -99,31 +99,36 @@ extension Date: ScalarRepresentable {
...
+    }
+}
+
+private extension TimeInterval {
+    /// Separates into integral and fractional, then round fractional to precision digits Int
+    /// - returns: Tuple of integral part and converted fractional part
+    func separateFractionalSecond(with precision: Int) -> (integral: TimeInterval, fractional: Int) {
+        var integral = 0.0
+        let fractional = modf(self, &integral)
+        let radix = pow(10.0, TimeInterval(precision))
```
jpsim: Nit: `TimeInterval(precision)` isn't a time interval

```suggestion
        let radix = pow(10.0, Double(precision))
```

Sources/Yams/Representer.swift
```diff
@@ -99,31 +99,36 @@ extension Date: ScalarRepresentable {
...
+        let dateWithoutNanosecond = Date(timeIntervalSinceReferenceDate: integral)
+        return iso8601WithoutZFormatter.string(from: dateWithoutNanosecond) +
+            String(format: ".%09d", nanosecond).trimmingCharacters(in: characterSetZero) + "Z"
+    }
+}
+
+private extension TimeInterval {
+    /// Separates into integral and fractional, then round fractional to precision digits Int
+    /// - returns: Tuple of integral part and converted fractional part
+    func separateFractionalSecond(with precision: Int) -> (integral: TimeInterval, fractional: Int) {
```
jpsim: Nit, this would make the call sites clearer:

```suggestion
    func separateFractionalSecond(withPrecision precision: Int) -> (integral: TimeInterval, fractional: Int) {
```

Sources/Yams/Representer.swift
```diff
@@ -99,31 +99,36 @@ extension Date: ScalarRepresentable {
...
+        guard nanosecond != 0 else { return iso8601Formatter.string(from: self) }
+
+        let dateWithoutNanosecond = Date(timeIntervalSinceReferenceDate: integral)
+        return iso8601WithoutZFormatter.string(from: dateWithoutNanosecond) +
+            String(format: ".%09d", nanosecond).trimmingCharacters(in: characterSetZero) + "Z"
+    }
+}
+
+private extension TimeInterval {
+    /// Separates into integral and fractional, then round fractional to precision digits Int
```
jpsim: Suggestion:

```suggestion
    /// Separates the time interval into integral and fractional components, then rounds the `fractional` component to `precision` number of digits.
```

Tests/YamsTests/TestHelper.swift
norio-nomura: Is the Swift optimizer smart enough to eliminate the code that adds `0` if `fraction` is `nil`?

norio-nomura: I changed to following:
```swift
    return fraction.map(date.addingTimeInterval) ?? date
```

Sources/Yams/Constructor.swift
norio-nomura: I changed it to the proposed code which changed the following part.
```swift
        return datecomponents.date.map { nanoseconds.map($0.addingTimeInterval) ?? $0 }
```

norio-nomura:
Thanks for reviews! 🙏 
I believe that I fixed everything pointed out from you!

Sources/Yams/Representer.swift
```diff
@@ -99,31 +99,36 @@ extension Date: ScalarRepresentable {
...
+}
+
+private extension TimeInterval {
+    /// Separates the time interval into integral and fractional components, then rounds the `fractional` component to `precision` number of digits.
+    /// - returns: Tuple of integral part and converted fractional part
+    func separateFractionalSecond(withPrecision precision: Int) -> (integral: TimeInterval, fractional: Int) {
+        var integral = 0.0
+        let fractional = modf(self, &integral)
+        let radix = pow(10.0, Double(precision))
+        let rounded = Int((fractional * radix).rounded())
```
norio-nomura: P.S

It should be truncated (`.rounded(.down)`?) instead of `rounded()`.

[The Data Formatting Guide](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/DataFormatting/Articles/dfDateFormatting10_4.html#//apple_ref/doc/uid/TP40002369-SW13) said:
> The format string uses the format patterns from the Unicode Technical Standard #35. 

And [the Unicode Technical Standard #35](http://www.unicode.org/reports/tr35/tr35-31/tr35-dates.html#Date_Field_Symbol_Table) said:

|Field|Sym.|No.|Example|Description|
|-----|----|---|-------|-----------|
|second|S|1..n|3456|Fractional Second - **truncates** (like other time fields) to the count of letters. (example shows display using pattern SSSS for seconds value 12.34567)|

`DateFormatter` rounds sub-millisecond and also caused # 192.
I filed this behavior and document inconsistency as a bug: https://bugs.swift.org/browse/SR-11049

For now, this implementation rounds sub-millisecond the same as `DateFormatter`, but this may have to be fixed.

PR #196, opened 2019-06-30
Add make-xcframework.sh script

jpsim:
We can't run this on CI yet because Azure Pipelines still doesn't provide images with Xcode 11 betas installed. See https://github.com/microsoft/azure-pipelines-image-generation/issues/1028

To run this locally, run `DEVELOPER_DIR=/Applications/Xcode-11.app/Contents/Developer ./make-xcframework.sh`.

This produces an XCFramework in `build/Yams.xcframework`.

### DIFF
--- a/make-xcframework.sh
+++ b/make-xcframework.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -e
+set -o pipefail
+set -u
+
+name=Yams
+project=$name.xcodeproj
+output=build/$name.xcframework
+for sdk in watchos watchsimulator iphoneos iphonesimulator appletvos appletvsimulator macosx; do
+  xcodebuild \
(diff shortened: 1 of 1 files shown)


jpsim:
Unfortunately this doesn't work with the SwiftPM-generated Xcode project because of two bugs:

* FB6395100: Running 'xcodebuild archive' on a SwiftPM product attempts to archive its test targets
* FB6395442: Running 'xcodebuild archive' on a SwiftPM product produces different results than an equivalent .xcodeproj

jpsim:
This seems to work now, command completes successfully with Xcode 12.2, but I'm not sure anyone actually _needs or wants_ this, so I'll close. We can integrate this into our release process if there's a demand for it.

PR #198, opened 2019-07-06
Apply `.singleQuoted` on representing `Node.Scalar` from `String` if `Resolver.default` resolves that to other than `.str`.

### LINKED ISSUE #197: Yams.dump dumps numerical String values as Number
this seems to be similar to https://github.com/jpsim/Yams/issues/174 though i'm not sure it's the same issue.

```swift
let dict = ["1.0": "2."]
print(try! Yams.dump(object: dict))
```

expected output: `'1.0': '2.'`
actual output: `1.0: 2.`

maybe this is an accepted limitation?

if so, i'd love to know if there is a sane way to correctly dump dicts of an unknown structure. i can only think of deep-mapping them to Nodes and setting styles based on the Strings' contents...

norio-nomura:
Fixes #197

### DIFF
--- a/Sources/Yams/Representer.swift
+++ b/Sources/Yams/Representer.swift
@@ -228,8 +228,8 @@ extension URL: ScalarRepresentable {
 extension String: ScalarRepresentable {
     /// This value's `Node.scalar` representation.
     public func represented() -> Node.Scalar {
-        let node = Node.Scalar.init(self, .init(.implicit, .default))
-        return node.resolvedTag.name == .str ? node : .init(self, .init(.str), .singleQuoted)
+        let scalar = Node.Scalar(self)
+        return scalar.resolvedTag.name == .str ? scalar : .init(self, Tag(.str), .singleQuoted)
     }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/Yams/Representer.swift
```diff
@@ -230,7 +228,8 @@ extension URL: ScalarRepresentable {
 extension String: ScalarRepresentable {
     /// This value's `Node.scalar` representation.
     public func represented() -> Node.Scalar {
-        return .init(self)
+        let scalar = Node.Scalar(self)
+        return scalar.resolvedTag.name == .str ? scalar : .init(self, Tag(.str), .singleQuoted)
```
norio-nomura: The performance impact can be almost ignored since `resolvedTag` was previously called in `serializeScalar(_:)` on serializing.
https://github.com/jpsim/Yams/blob/41c33b6f96/Sources/Yams/Emitter.swift#L413

norio-nomura:
> Should we backport this to a 2.0.x release without the breaking changes?

I'll branch `yams-2.0-branch` from 66946171bbf0fd178a98bde895f19a5f8128a416 and open cherry-picking PR against that

PR #200, opened 2019-07-07
Add Key coding strategy

norio-nomura:
It is a rewrite of #92.

#### `YAMLEncoder`
- Add `KeyEncodingStrategy` with cases: `useDefaultKeys`, `convertToSnakeCase` and `custom(_:)`
- Copy `Options` from `Emitter` with adding `keyEncodingStrategy`

#### `YAMLDecoder`
- Add `KeyDecodingStrategy` with cases: `useDefaultKeys`, `useSnakeCasedKeys`, `useCustomizedKeys(_:)` and `custom(_:)`
- Add `Options` having `keyDecodingStrategy`
- Change behaviors of `allKeys`:
  - Do not include duplicate keys
  - The order of the keys is no longer the same as in YAML

#### Difference between `JSONEncoder` and `YAMLEncoder` of `KeyEncodingStrategy`
The behavior is compatible with `JSONEncoder. KeyEncodingStrategy`

#### Difference between `JSONDecoder` and `YAMLDecoder` of `KeyDecodingStrategy`
- Drop `convertFromSnakeCase`, because that has a problem: [SR-6629](https://bugs.swift.org/browse/SR-6629)
- Add `useSnakeCasedKeys`  
  That uses snake cased `CodingKeys` for accessing YAML contents.
  *same with https://github.com/apple/swift/pull/14039*
  This replaces `convertFromSnakeCase`.
- Add `useCustomizedKeys((_ codingPath: [CodingKey]) -> CodingKey)`
  That uses custom converted `CodingKeys` for accessing YAML contents.
- `custom((_ codingPath: [CodingKey]) -> CodingKey)` is compatible with `JSONDecoder`'s one

### DIFF
--- a/Sources/Yams/Decoder.swift
+++ b/Sources/Yams/Decoder.swift
@@ -11,6 +11,9 @@ import Foundation
 /// `Codable`-style `Decoder` that can be used to decode a `Decodable` type from a given `String` and optional
 /// user info mapping. Similar to `Foundation.JSONDecoder`.
 public class YAMLDecoder {
+    /// Options to use when decoding from YAML.
+    public var options = Options()
+
     /// Creates a `YAMLDecoder` instance.
     ///
     /// - parameter encoding: Encoding, `.default` if omitted.
@@ -46,8 +49,8 @@ public class YAMLDecoder {
 
     /// Encoding
     public var encoding: Parser.Encoding
-    public var options = Options()
 
+    /// Configuration options to use when decoding YAML.
     public struct Options {
         /// The strategy to use for decoding keys. Defaults to `.useDefaultKeys`.
         public var keyDecodingStrategy: KeyDecodingStrategy = .useDefaultKeys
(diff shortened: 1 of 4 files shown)


doofyus:
Hi, what's the status with this one and can i be of help in moving it forward?
(`KeyDecodingStrategy` seems like an essential part, as Swift is commonly written in _camelCase_ vs yaml usually in _snake_case_)

beijingrice:
I view the webpage of WWDC2019, and found it’s lib on Github. Is that you found me by Github? My English maybe gets some error. :-)

beijingrice:
But I think that’s very good because everyone can link this lib to them’s Xcode project and that my our coding more easy.

frugoman:
Why was this closed? what's needed for this to be reopened and added to the lib? it's pretty essential

jpsim:
This is definitely something I'd like to see, not sure where @norio-nomura left it off. CI was failing but it seems that Microsoft Azure Pipelines hasn't kept the results from these runs.

PR #208, opened 2019-10-12
Yams: explicitly cast value for enums

compnerd:
Enums on Windows are different from Linux and macOS.  Add explicit casts
for the values to be constructed.  This is needed for building YAMS on
Windows.


jpsim:
Is this tracked by an SR at bugs.swift.org? It'd be good to link to that if so.

Thanks for the PR!

PR #210, opened 2019-10-12
CYaml: add explicit modulemap

compnerd:
Yams uses CYaml which requires a module.  s-p-m will implicitly
construct the modulemap for the module, but provide an explicit one for
better control and to enable building without s-p-m.

### DIFF
--- a/Sources/CYaml/include/module.modulemap
+++ b/Sources/CYaml/include/module.modulemap
@@ -1,0 +1,3 @@
+module CYaml {
+  header "yaml.h"
+}


norio-nomura:
If we explicitly use `module.modulemap`, `CYaml.h` should no longer be needed.

compnerd:
Yeah, `CYaml.h` shouldn't be needed with this, but I wanted to get these changes merged in before I tried to extract this file.

compnerd:
I don't have merge access, could you merge this for me please @norio-nomura?

PR #211, opened 2019-10-12
build: add CMake based build

compnerd:
This adds a CMake (3.15.1) based build to enable building Yams on
platforms where s-p-m does not yet function.  It also enables
cross-compilation of Yams.  This is sufficient to be able to build for
Windows with the following command line:

```
cmake -H build\debug -G Ninja -DFoundation_DIR=... -DCMAKE_Swift_COMPILER=...
cmake --build build\debug
```

### DIFF
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -1,0 +1,37 @@
+
+# Returns the current achitecture name in a variable
+#
+# Usage:
+#   get_swift_host_arch(result_var_name)
+#
+# If the current architecture is supported by Swift, sets ${result_var_name}
+# with the sanitized host architecture name derived from CMAKE_SYSTEM_PROCESSOR.
+function(get_swift_host_arch result_var_name)
+  if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
+    set("${result_var_name}" "x86_64" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
+    set("${result_var_name}" "aarch64" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "ppc64")
+    set("${result_var_name}" "powerpc64" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "ppc64le")
+    set("${result_var_name}" "powerpc64le" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "s390x")
+    set("${result_var_name}" "s390x" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "armv6l")
+    set("${result_var_name}" "armv6" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "armv7-a")
+    set("${result_var_name}" "armv7" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "armv7l")
+    set("${result_var_name}" "armv7" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "AMD64")
+    set("${result_var_name}" "x86_64" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "IA64")
+    set("${result_var_name}" "itanium" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86")
+    set("${result_var_name}" "i686" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "i686")
+    set("${result_var_name}" "i686" PARENT_SCOPE)
+  else()
+    message(FATAL_ERROR "Unrecognized architecture on host system: ${CMAKE_SYSTEM_PROCESSOR}")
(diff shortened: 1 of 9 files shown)


norio-nomura:
Where can I find instructions on how to provide FoundationConfig.cmake for macOS?
```terminal.sh-session
$ cmake -G Ninja -DCMAKE_Swift_FLAGS="-sdk $(xcrun --sdk macosx --show-sdk-path)"
CMake Warning:
  No source or binary directory provided.  Both will be assumed to be the
  same as the current working directory, but note that this warning will
  become a fatal error in future CMake releases.


CMake Error at CMakeLists.txt:11 (find_package):
  Could not find a package configuration file provided by "Foundation" with
  any of the following names:

    FoundationConfig.cmake
    foundation-config.cmake

  Add the installation prefix of "Foundation" to CMAKE_PREFIX_PATH or set
  "Foundation_DIR" to a directory containing one of the above files.  If
  "Foundation" provides a separate development package or SDK, be sure it has
  been installed.


-- Configuring incomplete, errors occurred!
See also "/Users/norio/github/swift-dev/SourceKitten/Carthage/Checkouts/Yams/CMakeFiles/CMakeOutput.log".
See also "/Users/norio/github/swift-dev/SourceKitten/Carthage/Checkouts/Yams/CMakeFiles/CMakeError.log".
```

compnerd:
Ah; that is dependent on apple/swift-corelibs-foundation#2524; you can add `-DFoundation_DIR=...` after that gets merged (I'm looking into the LLDB failures still before I can merge that).

norio-nomura:
I'm trying to build https://github.com/apple/swift-corelibs-foundation/pull/2524 on a local VM with ubuntu 16.04. However, neither `FoundationConfig.cmake` nor `foundation-config.cmake` is generated under build directories or working copy directories.

jpsim:
This is lovely, @compnerd! Once apple/swift-corelibs-foundation#2524 lands and it becomes possible to build Yams on Windows, we should add Windows to our CI pipeline.

We should also advertise CMake and Windows support in `CHANGELOG.md` and `README.md`.

compnerd:
@jpsim @norio-nomura - okay that change has been merged.  Can we get this tested and merged?  Then we can update the readme and CI configuration.

jpsim:
Is `FoundationConfig.cmake` something we should add to Yams, or does it indicate that corelibs foundation should expose something similar?

norio-nomura:
I think we should basically use what s-c-f created.  However, since s-c-f currently does not support building with cmake on macOS, I think that it may be possible to attach that  macOS to simplify CI.

norio-nomura:
I'm sorry, I was closing the wrong operation. 🙇‍♂️

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,37 @@
+
+# Returns the current achitecture name in a variable
+#
+# Usage:
+#   get_swift_host_arch(result_var_name)
+#
+# If the current architecture is supported by Swift, sets ${result_var_name}
+# with the sanitized host architecture name derived from CMAKE_SYSTEM_PROCESSOR.
+function(get_swift_host_arch result_var_name)
```
jpsim: It feels odd that this would need to live in Yams. Is it possible for this logic to live in Swift or CMake?

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,37 @@
+
+# Returns the current achitecture name in a variable
+#
+# Usage:
+#   get_swift_host_arch(result_var_name)
+#
+# If the current architecture is supported by Swift, sets ${result_var_name}
```
jpsim: What actually reads `result_var_name`? I don't see it referenced anywhere else in these files. It'd be good to document this.

cmake/modules/SwiftSupport.cmake
compnerd: This is something specific to the layout that is application specific.  This perhaps belongs in some shared swift repository, but, for now, this is replicated.

cmake/modules/SwiftSupport.cmake
compnerd: Its the output variable, not an input variable.

jpsim: I understand that, but what reads this output?

compnerd:
Yeah, I would like to add s-p-m to the nightlies, but that is being blocked by some updates to the CI hosts (hopefully that will be done soon).  I just tend to prefer splitting up the docs and the build infrastructure changes.  But, I would certainly like to add documentation on how to build this on Windows.

cmake/modules/SwiftSupport.cmake
compnerd: The caller of the function will read the output (which can be anything really, its completely opaque to the function).

compnerd:
Anything that I can help with to get this merged?

jpsim:
Just what I said in my last comment, regarding changelog, readme and `.gitignore`:

* Should we add anything to the `.gitignore` file, like CMake or Ninja related artifacts?
* If this enables Yams to be built on Windows, can we say as much in the changelog and in the readme with instructions on how to do so?

compnerd:
> Just what I said in my last comment, regarding changelog, readme and `.gitignore`:
> 
> * Should we add anything to the `.gitignore` file, like CMake or Ninja related artifacts?

Nope, the build artifacts are placed where the user requests it, and in-tree builds are discouraged.

> * If this enables Yams to be built on Windows, can we say as much in the changelog and in the readme with instructions on how to do so?

Sure, I can do that.  Would you want that to be part of this PR or a different one?

jpsim:
At the very least changelog should be in the PR introducing the change, but documentation in the readme can come later if that'll take more time.

compnerd:
@jpsim does that seem reasonable enough for the changelog and readme?

jpsim:
Great, thanks for adding that. I still think it'd be good to point to what steps need to be taken before being able to run those CMake commands, but it's a good start.

PR #220, opened 2019-11-03
[GitHub Actions] Add xcodebuild workflow

norio-nomura:
It contains Codecov step that will be enabled by setting `secrets.CODECOV_TOKEN`.


norio-nomura:
`CODECOV_TOKEN` can be get from https://codecov.io/gh/jpsim/Yams.
`CODECOV_TOKEN` can be set by following [the instruction](https://help.github.com/en/github/automating-your-workflow-with-github-actions/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables).

PR #225, opened 2019-11-03
[GitHub Actions] Add SwiftLint workflow

norio-nomura:
- Uses `norio-nomura/action-swiftlint`
- It won't work with PR from forked repository.


jpsim:
> It won't work with PR from forked repository.

Looks like this is explicitly restrained in the action. Does it break if we remove that condition?

norio-nomura:
> > It won't work with PR from forked repository.
> 
> Looks like this is explicitly restrained in the action. Does it break if we remove that condition?

Removing `if:` will cause "server error status: 403" on PR from forked repository.

PR #228, opened 2019-11-03
Use default protocol implementations where possible

jpsim:
From what I can tell, these custom implementations don't improve on the default ones in any way.

### DIFF
--- a/Sources/Yams/Node.Sequence.swift
+++ b/Sources/Yams/Node.Sequence.swift
@@ -114,11 +114,6 @@ extension Node.Sequence: MutableCollection {
     }
 
     /// :nodoc:
-    public func index(after index: Index) -> Index {
-        return nodes.index(after: index)
-    }
-
-    /// :nodoc:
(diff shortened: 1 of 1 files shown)


Sources/Yams/Node.Sequence.swift
```diff
@@ -146,24 +141,7 @@ extension Node.Sequence: MutableCollection {
...
-extension Node.Sequence: RandomAccessCollection {
-    // BidirectionalCollection
-    /// :nodoc:
-    public func index(before index: Index) -> Index {
-        return nodes.index(before: index)
-    }
-
-    // RandomAccessCollection
-    /// :nodoc:
-    public func index(_ index: Index, offsetBy num: Int) -> Index {
```
norio-nomura: It seems that only this func is not used in the code coverage before this PR. 
The other func to be deleted was used in the previous tests, so it seems to be no problem.

jpsim:
Going to close this in case it has perf implications.

PR #231, opened 2019-11-10
Update CI

norio-nomura:


jpsim (approved):
Thanks so much for doing this. Sucks that GitHub pulled support for Xcode 9/10 without warning.

PR #234, opened 2019-11-27
Added indentation tests for the emitter

acecilia:


Tests/YamsTests/EmitterTests.swift
```diff
@@ -48,6 +48,28 @@ class EmitterTests: XCTestCase {
...
+
+            """
+        }
+        XCTAssertEqual(try Yams.serialize(node: node), expected(2))
+        XCTAssertEqual(try Yams.serialize(node: node, indent: -2), expected(2))
+        XCTAssertEqual(try Yams.serialize(node: node, indent: -1), expected(2))
+        XCTAssertEqual(try Yams.serialize(node: node, indent: 0), expected(2))
+        XCTAssertEqual(try Yams.serialize(node: node, indent: 4), expected(4))
+        XCTAssertEqual(try Yams.serialize(node: node, indent: 9), expected(9))
+        XCTAssertEqual(try Yams.serialize(node: node, indent: 10), expected(2))
```
jpsim: Some of these results are surprising to me. Is this a bug?

acecilia: No, no bug.

I can give you some more context about why I end up opening this PR. After I saw that the arrays where not being indented when encoding with the `YAMLEncoder`, I thought it was a bug. I cloned the repo and did some tests, and apparently all is working as expected: the YAML specification allows arrays without indentation, which is what Yams is producing. After looking into it I had the tests done, so I decided to PR them :)

The indentation level has to be between 2 and 10 (hardcoded in the C logic), as you can see in this tests.

jpsim: Ah makes sense. Thanks for writing tests! ❤️

PR #238, opened 2020-01-07
Fix `YAMLDecoder` to support merging anchors

brentleyjones:
Currently `YAMLDecoder` uses `Resolver.basic` instead of `Resolver.default`, since it doesn't need to resolve types. This has the side effect of not including `.merge` as an option, making `decode` fail to see anchored properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@
   [Norio Nomura](https://github.com/norio-nomura)
   [#192](https://github.com/jpsim/Yams/issues/192)
 
+* Fix `YAMLDecoder` to support merging anchors.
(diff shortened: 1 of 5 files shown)


brentleyjones:
I should have opened an issue first. This might not actually fix the issue. Regardless I should probably add tests, changelog, etc.

jpsim:
Thanks for the change! It looks right to me, but I'd like to see the tests and changelog entry as you mentioned. No need to file a separate ticket for the issue you're trying to resolve.

brentleyjones:
@jpsim I fixed the fix, added some tests, and updated the changelog.

jpsim:
Thanks!

PR #254, opened 2020-05-14
build: simplify Foundation handling

compnerd:
Enable building against an uninstalled copy of Foundation but do not
require it.  This allows for a simpler build system.  If the user wishes
to build against an out-of-tree Foundation, they can simply specify
`Foundation_DIR` and `dispatch_DIR` when configuring and CMake will
handle the rest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,18 +8,8 @@ project(Yams
 
 option(BUILD_SHARED_LIBS "build shared libraries" ON)
 
-if (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for the PR. Would you mind rebasing? It looks like CI choked on the commit sha, and I'm hoping that https://github.com/jpsim/Yams/pull/257 fixes that.

compnerd:
Sure!

compnerd:
ugh, right, Foundation is a framework on Darwin.

jpsim:
Looks great, thank you!

PR #255, opened 2020-05-14
build: use a single top-level Swift module directory

compnerd:
Rather than have a swift module directory that you need to find for each
library, use a single global swift module directory similar to the
runtime output directory.  This will be useful for a follow up change to
enable tests.


jpsim:
Thanks for the PR!

PR #259, opened 2020-05-18
Yams: remove conditional cases for Windows

compnerd:
Unify the non-Windows and Windows paths.  The full codebase now builds
without any special case for Windows.


jpsim:
Lovely, thanks!

PR #260, opened 2020-05-20
Fix use of pow in separateFractionalSecond

asuhan:
Fixes the `Yams/Representer.swift:130:21: error: static member 'pow' cannot be used on instance of type 'Double'` we hit.

### DIFF
--- a/Sources/Yams/Representer.swift
+++ b/Sources/Yams/Representer.swift
@@ -127,7 +127,11 @@ private extension TimeInterval {
     func separateFractionalSecond(withPrecision precision: Int) -> (integral: TimeInterval, fractional: Int) {
         var integral = 0.0
         let fractional = modf(self, &integral)
+#if os(Linux)
+        let radix = Glibc.pow(10.0, Double(precision))
+#else
         let radix = pow(10.0, Double(precision))
+#endif
         let rounded = Int((fractional * radix).rounded())
         let quotient = rounded / Int(radix)
         return quotient != 0 ? // carry-up?


jpsim:
How are you hurting that error? Is that a compile time error or runtime? This shouldn’t be happening as best I can tell.

asuhan:
@jpsim It's a compile error, we hit it when building the Swift toolchain on Ubuntu 18.04.

jpsim:
> It's a compile error, we hit it when building the Swift toolchain on Ubuntu 18.04.

That's odd. Can you please share steps to reproduce the issue? I'm not seeing that on Ubuntu 18.04 with Swift 5.2.2 (also tested with Swift 5.2.3):

```
$ docker run -it --rm swift:5.2.2 bash 
root@d25a41d005ed:/# cat /etc/issue
Ubuntu 18.04.4 LTS \n \l

root@d25a41d005ed:/# git clone https://github.com/jpsim/Yams.git
Cloning into 'Yams'...
remote: Enumerating objects: 11, done.
remote: Counting objects: 100% (11/11), done.
remote: Compressing objects: 100% (11/11), done.
remote: Total 4869 (delta 0), reused 6 (delta 0), pack-reused 4858
Receiving objects: 100% (4869/4869), 6.59 MiB | 6.84 MiB/s, done.
Resolving deltas: 100% (2865/2865), done.
root@d25a41d005ed:/# cd Yams
root@d25a41d005ed:/Yams# swift build
[22/22] Merging module Yams
root@d25a41d005ed:/Yams# echo $?
0
root@d25a41d005ed:/Yams#
```

asuhan:
I'm building from https://github.com/apple/swift/tree/tensorflow-merge. The closest toolchain would be https://github.com/apple/swift/releases/tag/swift-DEVELOPMENT-SNAPSHOT-2020-05-18-a, I think.

asuhan:
Looks like we're not getting the right `pow`, it works if I explicitly qualify it with `Glibc`. @compnerd might now more about this.

compnerd:
@asuhan  if you qualify, you need to qualify for multiple targets, but of course it would work when qualified.

The problem here is that pow is not a real function.  The functions are powf, powl, and powll.  pow is a swift extension to add a type generic function a la C11.  The qualified name is exactly that and is why it works.

This might just be a race condition: import System/Glibc/MSVCT and it should work.  The typed function is just directly going to the underlying libc function and is cheaper though.

asuhan:
@compnerd I've already tried to import System/Glibc/MSVCT and didn't get it to work. If I convert to powl, I get:

```
yams/Sources/Yams/Representer.swift:137:32: error: cannot convert value of type 'Double' to expected argument type 'Float80'
        let radix = powl(10.0, Double(precision))
                               ^
                               Float80(         )
```

If I don't, I get:

```
swift-source/yams/Sources/Yams/Representer.swift:137:21: error: static member 'pow' cannot be used on instance of type 'Double'
        let radix = pow(10.0, Double(precision))
                    ^~~
                    Double.
```

jpsim:
@asuhan could you please file a bug at https://bugs.swift.org? I'd consider it a Swift bug if `pow(10.0, Double(1.0))` fails to compile on any platform.

asuhan:
Turns out the issue is specific to the TF branch: https://github.com/asuhan/Yams/pull/1. Sorry!

jpsim:
I’m glad you got to the bottom of this.

Issue #263, opened 2020-07-08
String representations of numbers can be decoded as numbers.


mattpolzin:
This feels tangential to https://github.com/jpsim/Yams/issues/174 but in fact is distinct and specifically related to Yams's `Decoder` implementation.

While string values may be quoted or (more often) not in YAML, the easiest way to differentiate a number from a string scalar containing only the textual representation of a number is to quote the string representation.

```yaml
number: 1234
string: '1234'
```

Currently, Yams will successfully decode the above YAML as the following type.

```swift
struct Funky: Codable {
  let number: Int
  let string: Int
}
```

This is arguably a problem for validity checking and definitely causes inconsistency in that re-encoding the same struct would result in different YAML.

```yaml
number: 1234
string: 1234
```

An additional problem arises when a type to be decoded might sometimes be a number and other times a string. For example, you may be working with an `AnyCodable` type (there are a few popular implementations of such a type) because you need to decode an arbitrary structure. These types tend to employ a fallback strategy of decoding that results in _trying_ to decode an integer or other numeric type and succeeding even for quoted YAML scalars as long as those string values happen to be valid string representations of the given numeric type.

PR #264, opened 2020-07-08
Fix and test string -> integer misstep upon decoding.

mattpolzin:
Fixes https://github.com/jpsim/Yams/issues/263 by only constructing integer and floating point values from scalars if those scalars are in the `plain` "flow" style. It additionally allows through scalars with no styling information (`any`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@
 
 ##### Bug Fixes
 
-* None.
+* String scalars containing numbers are no longer decoded as numbers.
+  [Matt Polzin](https://github.com/mattpolzin)
+  [#263](https://github.com/jpsim/Yams/issues/263)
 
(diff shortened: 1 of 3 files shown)


mattpolzin:
@jpsim apologies for the ping, but I'd love to get this fix applied (or officially declined). It's the only thing I currently know of that stops me from using Yams to ingest OpenAPI documentation in my a particular production application of OpenAPIKit.

jpsim:
Looks great, thank you for the ping!

I just updated the changelog entry position to reflect that it's landing in master, will merge when CI passes.

Thanks for the PR.

jpsim:
Do you need me to tag a release with this or can you point your use of Yams to master?

mattpolzin:
I can point at master. Thank you!

PR #273, opened 2020-08-31
Fix build with tensorflow toolchain

garymm:
Due to https://bugs.swift.org/browse/TF-1203, there's a build failure
without this patch:

```
error: static member 'pow' cannot be used on instance of type 'Double'
```

I believe this was the intent of 70140a180ae63292c5189815d35480269087c306, but it didn't seem to actually fix it.

### DIFF
--- a/Sources/Yams/Representer.swift
+++ b/Sources/Yams/Representer.swift
@@ -130,7 +130,13 @@ private extension TimeInterval {
     func separateFractionalSecond(withPrecision precision: Int) -> (integral: TimeInterval, fractional: Int) {
         var integral = 0.0
         let fractional = modf(self, &integral)
+        // Can't use `pow` free function due to https://bugs.swift.org/browse/TF-1203.
+        // TODO: Remove condition after that bug is fixed.
+        #if canImport(TensorFlow)
+        let radix = Double.pow(10.0, Double(precision))
+        #else
         let radix = pow(10.0, Double(precision))
+        #endif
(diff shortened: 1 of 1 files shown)


jpsim:
Thank you for filing (or at least finding) the source bug. I added a comment linking to it. Will merge when CI passes.

jpsim:
Ah, this is causing issues:

```

/__w/Yams/Yams/Sources/Yams/Representer.swift:134:28: error: type 'Double' has no member 'pow'
12
        let radix = Double.pow(10.0, Double(precision))
13
                    ~~~~~~ ^~~
```

jpsim:
@garymm could you please try the latest commit from this branch with your project to see if it works around the bug?

jpsim:
Also see this other discussion and fix for context: https://github.com/asuhan/Yams/pull/1

garymm:
No it doesn't work sadly. Yesterday I tried a bunch of different type conversions but I couldn't get anything to work except using `Double.pow()`. If you think it's impossible to make it work with both the official and Tensorflow toolchains, feel free to close this.

CC @dan-zheng @asuhan

jpsim:
Is there a compile-time check we can add to only use `Double.pow` when compiling against Swift for Tensorflow?

garymm:
This seems to work, but I don't know if it's acceptable?

jpsim:
Yes that's great, thank you! Will merge when CI passes.

PR #282, opened 2020-11-03
Add Windows CI job

jpsim:
Copied from https://github.com/compnerd/swift-com/blob/f6712a7/.github/workflows/swift.yml


jpsim:
@compnerd would you mind reviewing this to see if there’s anything you suggest I do differently?

compnerd (commented):
LGTM; interesting issue that you've uncovered, I wonder what's going on there.  As far as GitHub Actions go, this is the best that I've been able to come up with as well.

jpsim:
Thanks for looking this over!

PR #291, opened 2020-11-19
Update Xcode project to Swift 5.0

brennanMKE:
* updates Swift 4.2 to 5.0
* enables `CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER`
* updates Xcode version to 12.2
* updates CHANGELOG.md


jpsim:
Thanks for the PR! Do you actually need the Xcode project? I've removed it from SwiftLint a few weeks ago and it's something that I'd like to do with all my projects now, since Xcode has had good SwiftPM support in the last two major versions.

brennanMKE:
We still use Xcode projects since the binary artifacts support is not ready for SPM yet. That is the limit for us.

brennanMKE:
Below is what I have been following so that we can use pre-compiled dependencies which are accessible with a package registry. Part of it has been implemented while the rest may be in Xcode vNext.

* [Binary Proposal](https://github.com/FranzBusch/swift-evolution/pull/1)
* [Binary Dependencies](https://github.com/apple/swift-evolution/blob/master/proposals/0272-swiftpm-binary-dependencies.md)
* [Registry Service](https://forums.swift.org/t/swift-package-registry-service/37219)
* [Source Archive Dependencies](https://forums.swift.org/t/package-manager-source-archive-dependencies/38626/8)

jpsim:
I’m not following, we don’t offer binaries today, so why would removing the Xcode project have anything to do with that?

brennanMKE:
Our build system uses the Xcode project file which creates build products in binary form. Since SPM does not support this functionality we still use Xcode project files like they have been for years.

jpsim:
If you produce your own binary artifacts, then can you also produce your own Xcode project? For example, using `swift package generate-xcodeproj`?

brennanMKE:
This is not how we do it currently as I believe the generated identifiers are significant so they have to be persisted.

jpsim:
> This is not how we do it currently as I believe the generated identifiers are significant so they have to be persisted.

Can you please elaborate on this?

I'd really not rather keep the pbxproj file around at all, especially not to support some custom closed source integration that I can't confirm if I'm breaking at any point.

brennanMKE:
Since I've imported this code we may not need to keep the Xcode project file in the GitHub repo. If I do merge updates later I can simply persist the Xcode project file with the changes I have made. Normally I submit a PR back to a GitHub repo when I make changes to sure future merges don't require a patch, but I think this won't cause a lot more effort on my side.

jpsim:
I'll make a release with this but the Xcode project will be removed for the next release.

jpsim:
https://github.com/jpsim/Yams/releases/tag/4.0.3

PR #293, opened 2020-12-08
Add support for Apple Silicon in SwiftSupport.cmake

MaxDesiatov:
This matches the behavior of [a similar `elseif` branch in the Swift toolchain itself](https://github.com/apple/swift/blob/b80e4bf8f866be26a4108bd92f97b2a182662642/CMakeLists.txt#L594). I'm unable to make it work with `MATCHES` though, so I've added a plain `STREQUAL`, which is also consistent with the rest of the branches.


jpsim:
Thanks!

MaxDesiatov:
@jpsim would you be able to also tag a new patch release with this? The toolchain depends on a specific tag, and I'd rather just bump the tag than depend on `master`.

jpsim:
Released here: https://github.com/jpsim/Yams/releases/tag/4.0.2

MaxDesiatov:
Much appreciated! 🙏

jpsim:
Just out of curiosity, what are you building with Yams via CMake on Apple Silicon? Is this for `swift-driver`?

MaxDesiatov:
Yes, `swift-driver`. That's the only consumer of Yams via CMake, as far as as I'm aware 😄

PR #296, opened 2021-01-11
Expose the underlying Node's Mark on Decoder

brentleyjones:
This allows accessing source location during decoding.

### DIFF
--- a/Sources/Yams/Decoder.swift
+++ b/Sources/Yams/Decoder.swift
@@ -69,7 +69,7 @@ public class YAMLDecoder {
 
 private struct _Decoder: Decoder {
 
-    private let node: Node
+    fileprivate let node: Node
 
     init(referencing node: Node, userInfo: [CodingUserInfoKey: Any], codingPath: [CodingKey] = []) {
         self.node = node
@@ -341,6 +341,15 @@ extension URL: ScalarConstructible {
     /// - returns: An instance of `URL`, if one was successfully extracted from the scalar.
     public static func construct(from scalar: Node.Scalar) -> URL? {
         return URL(string: scalar.string)
+    }
+}
+
+// MARK: Decoder.mark
+
+extension Decoder {
+    /// The `Mark` for the underlying `Node` that has been decoded.
+    public var mark: Mark? {
+        return (self as? _Decoder)?.node.mark
     }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/Yams/Decoder.swift
```diff
@@ -344,6 +344,19 @@ extension URL: ScalarConstructible {
     }
 }
 
+// MARK: Decoder.node
```
jpsim, line 347: Nit:
```suggestion
// MARK: Decoder.mark
```

Sources/Yams/Decoder.swift
```diff
@@ -344,6 +344,19 @@ extension URL: ScalarConstructible {
...
+
+extension Decoder {
+    /// The `Mark` for the underlying `Node` that has been decoded.
+    public var mark: Mark? {
+        guard let decoder = self as? _Decoder else {
+            return nil
+        }
+
+        return decoder.node.mark
```
jpsim, line 356: Nit:

```suggestion
        return (self as? _Decoder)?.node.mark
```

jpsim (commented):
Seems like a very useful thing to have access to, thanks for the PR!

I just have two small nitpicky suggestions. Could you also please add a test and a changelog entry?

Sources/Yams/Decoder.swift
brentleyjones: :+1: Missed in a refactor.

Tests/YamsTests/EncoderTests.swift
```diff
@@ -348,6 +348,26 @@ class EncoderTests: XCTestCase { // swiftlint:disable:this type_body_length
...
+            init(from decoder: Decoder) throws {
+                let container =  try decoder.singleValueContainer()
+
+                self.value = try container.decode(String.self)
+                self.location = decoder.mark?.description ?? ""
+            }
+        }
+        let t = try YAMLDecoder().decode(Test.self, from: s)
+        XCTAssertEqual(t.value, "Value")
+        XCTAssertEqual(t.location, "1:1")
```
brentleyjones, line 368: I would have tested a mapping here, but #297 prevents the value from being what it should be.

jpsim: Can you rebase this PR after merging https://github.com/jpsim/Yams/pull/297 so you can add that test?

brentleyjones: :+1: I'll also add the linux change needed.

jpsim (approved):

PR #297, opened 2021-01-11
Fix mark for sequences and mappings

brentleyjones:
Instead of the start mark for the sequence or map we were capturing the start mark of the _next_ element.

### DIFF
--- a/Tests/YamsTests/MarkTests.swift
+++ b/Tests/YamsTests/MarkTests.swift
@@ -39,12 +39,31 @@ class MarkTests: XCTestCase {
                 "'identifier_name' and will be completely removed in a future release."
             ])
     }
+
+    func testMappingMarkIsCorrect() throws {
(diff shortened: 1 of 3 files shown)


Tests/YamsTests/MarkTests.swift
```diff
@@ -39,6 +39,24 @@ class MarkTests: XCTestCase {
                 "'identifier_name' and will be completely removed in a future release."
             ])
     }
+
+    func testMappingMarkIsCorrect() throws {
```
jpsim, line 43: Yams manually tracks tests for Linux in `allTests`, so could you please add this test in the `allTests` extension below?

jpsim (approved):
Fix looks good, thank you! Could you please add a changelog entry?

I've given you commit access so you can merge yourself once that's done and CI passes.

PR #300, opened 2021-03-06
Expose the CYaml library as an installable product

NSCoder:
Hi Maintainers 👋 !

Thank you for creating this swift library! I currently need to access the underlying CYaml library and I was hoping we could expose it as a product in the package.json file.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,8 @@ import PackageDescription
 let package = Package(
     name: "Yams",
     products: [
-        .library(name: "Yams", targets: ["Yams"])
+        .library(name: "Yams", targets: ["Yams"]),
+        .library(name: "CYaml", targets: ["CYaml"])
     ],
     targets: [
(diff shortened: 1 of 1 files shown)


jpsim:
Could you please elaborate on what your goals are for doing this? Is there something we can add or change to the Yams interface to enable your goals?

I have a few concerns right off the bat:

1. If we expose CYaml as a public target, then we’ll need to consider its interface determining semantic versioning.
2. If ever we want to write a new yaml parser implementation in Swift and no longer need libyaml as a dependency, it’d be nice to just remove it without having to worry about breaking some users.

jpsim:
Let me know if ever you'd like to revisit this based on the concerns I raised in my last comment. For now I'll be closing this.

NSCoder:
Thank you for closing it @jpsim, you are right about your concerns and in that light this code does not longer make sense

PR #304, opened 2021-03-22
[CI] Debug Windows CI job

jpsim:
Failing [here](https://github.com/jpsim/Yams/pull/302/checks?check_run_id=2162561317) for seemingly unrelated reasons.

PR #305, opened 2021-03-22
build: build libyaml as a shared module on Windows

compnerd:
The libyaml setup is currently configured improperly, not selecting
either shared nor static builds, resulting in libyaml being built
assuming that it is a user of libyaml, which is incorrect.  Adjust the
build to build libyaml as a shared library.

This change is not exactly precise, over generalizing the application of
the `YAML_DECLARE_EXPORT` to avoid bumping the minimum version of SPM to
5.4, which is the first version to include support for
`.platform(.windows)`.  This should not impact the other targets, as the
macro is not consulted for other targets.

This should allow us to enable testing on Windows again which was
disabled with #304.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,7 +7,7 @@ let package = Package(
         .library(name: "Yams", targets: ["Yams"])
     ],
     targets: [
-        .target(name: "CYaml"),
+        .target(name: "CYaml", cSettings: [.define("YAML_DECLARE_EXPORT")]),
         .target(name: "Yams", dependencies: ["CYaml"]),
         .testTarget(name: "YamsTests", dependencies: ["Yams"])
     ]


jpsim (approved):
👍 This seems to match what's in the `CMakeLists.txt` for libyaml: https://github.com/yaml/libyaml/blob/acd6f6f014c25e46363e718381e0b35205df2d83/CMakeLists.txt#L64-L66

jpsim:
A big thank you for the investigation and solution on this @compnerd. It's still surprising to me that this would have broken seemingly only after an update to the Windows GitHub Action virtual environment.

compnerd:
You're welcome :) BTW, I think that this probably broke with the fix for the target check (that is, `WIN32` vs `_WIN32` which is the officially sanctioned macro to identify a Windows environment (https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160)

PR #310, opened 2021-03-23
Move Yams Bazel targets to the top level

tinder-maxwellelliott:
Currently developers cannot access these nested targets, moving to the
top level will allow them to consume them in their bazel projects


tinder-maxwellelliott:
@Keith Does this look correct to you? Thanks

.bazelrc
```diff
@@ -1,4 +1,3 @@
 # Force the tests to not run in sandboxed mode, since it can introduce
 # errors and flakes.
 test --spawn_strategy=local
-build --macos_minimum_os=10.9
```
keith, line 4: not to add more thrash, but I guess you would probably still prefer this here, unless setting them on the build test targets is enough, just in case people run the tests locally?

BUILD
```diff
@@ -0,0 +1,31 @@
...
+    "swift_library"
+)
+
+cc_library(
+    name = "CYaml",
+    srcs = glob(["Sources/CYaml/src/*.c", "Sources/CYaml/src/*.h"]),
+    hdrs = ["Sources/CYaml/include/yaml.h"],
+    includes = ["Sources/CYaml/include"],
+    visibility = [
+        "//visibility:public",
```
keith, line 16:
```suggestion
        "//Tests:__subpackages__",
```


probably still only want this?

keith (approved):
tested locally w/ our project and this fixes it

Tests/BUILD
```diff
@@ -9,11 +9,18 @@ config_setting(
     constraint_values = ["@platforms//os:linux"],
 )
 
+genrule(
+    name = "LinuxMain",
+    srcs = ["LinuxMain.swift"],
+    outs = ["main.swift"],
+    cmd = "cp $< $@"
+)
```
tinder-maxwellelliott: @Keith without this change Linux builds do not work, how did you get this working on other projects?

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
+* Update Bazel config to allow targets to be directly consumed.
```
jpsim, line 9: Sorry I know it's an annoyingly arcane syntax, but this needs two trailing spaces for Markdown to translate it to a line break.

```suggestion
* Update Bazel config to allow targets to be directly consumed.  
```

Tests/BUILD
keith: I'd have to debug a bit but we use this macro which definitely outputs it as LinuxMain.swift:


```bzl
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library", "swift_test")

def swiftpm_test(
        name,
        srcs,
        deps,
        data = [],
        test_tags = []):
    """Emulate Swift Package Manager's test target behavior

    Args:
        name: The name of the test target
        srcs: A list of source files to compile for the tests
        deps: A list of targets the test module depends on
        data: Any resources that should be made available at test time
        test_tags: Additional tags to pass to the test target
    """
    library_name = name + ".library"
    swift_library(
        name = library_name,
        srcs = srcs,
        data = data,
        deps = deps,
        module_name = name,
    )

    rule_name = "{}.linuxmain".format(name)
    test_srcs = [rule_name]
    native.genrule(
        name = rule_name,
        srcs = srcs,
        cmd = "./$(location //tools:linux_main_generator) {} $(SRCS) > $(OUTS)".format(name),
        outs = ["Tests/LinuxMain.swift"],
        exec_tools = ["//tools:linux_main_generator"],
        message = "Generating LinuxMain.swift",
    )

    swift_test(
        name = name,
        srcs = select({"//:linux": test_srcs, "//conditions:default": []}),
        deps = [library_name],
        tags = test_tags,
        size = "small",
    )
```

jpsim:
I'm happy merging this as-is @tinder-maxwellelliott unless you want to look further into the `LinuxMain.swift` issue.

tinder-maxwellelliott:
> I'm happy merging this as-is @tinder-maxwellelliott unless you want to look further into the `LinuxMain.swift` issue.

Lets merge it in, the `LinuxMain.swift` move is an optimization

jpsim:
@keith can you confirm this is usable externally before I cut a release?

keith:
Still seeing some issues, I'll submit a PR

keith:
https://github.com/jpsim/Yams/pull/312

keith:
Ok I understand the LinuxMain.swift issue now, the difference between our macro I posted above and this use case is we create a separate swift_library under the hood with all sources, and then the swift_test target only contains the LinuxMain.swift file, this is also what SwiftPM does [under the hood](https://github.com/apple/swift-package-manager/blob/61a369bf07dd030d25dca1b2628a0c4104e7ba4b/Sources/PackageLoading/PackageBuilder.swift#L1143-L1146), and in that case, where only a single file is passed, swiftc treats it as the main file no matter what the name is (unless -parse-as-library is passed):

```
% swiftc /tmp/foo.swift
% swiftc /tmp/foo.swift /tmp/bar.swift
/tmp/foo.swift:1:1: error: expressions are not allowed at the top level
print("hi")
^
```

The reason renaming works is because in that case it goes through a special codepath in swiftc https://github.com/apple/swift/blob/8ae1b768f1dc6e613f88878569f8024e350acb7d/lib/Frontend/Frontend.cpp#L857-L862

So I think in general we can leave this as is, but if we wanted to eliminate it I would create a separate swift_library target and depend on that from the swift_test (which would also fix the import warning we see today)

PR #311, opened 2021-03-23
Add swift_binary build target

keith:
This target is a bit different from the others that already exist
because it supports building on macOS and Linux, but it also has some
different semantics because of that so I think it's worth testing here.


keith:
Hoping to see a failure here first

keith:
This wasn't a good enough repro for the case we see an issue with. Actual fixes are in https://github.com/jpsim/Yams/pull/312

PR #312, opened 2021-03-23
Update BUILD configuration

keith:
This has 3 important fixes from Lyft's internal config:

1. The `cc_library` is now tagged with `swift_module` which is a special
   tag that tells rules_swift to generate a module map for the library.
   This is likely why sandboxing had to be disabled.
2. This adds `-fPIC` to the `cc_library` compilation. This is required
   for C libraries that Swift depends on, but there's an outstanding bug
   in bazel where this isn't handled correctly when you use the target
   through the host transition.
3. Force CYaml to be statically linked

This also formats this file using buildifier

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,1 @@
-# Force the tests to not run in sandboxed mode, since it can introduce
-# errors and flakes.
 build --macos_minimum_os=10.9
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,4 +1,3 @@
 # Force the tests to not run in sandboxed mode, since it can introduce
 # errors and flakes.
-test --spawn_strategy=local
 build --macos_minimum_os=10.9
--- a/BUILD
+++ b/BUILD
@@ -1,31 +1,26 @@
-load(
-    "@rules_cc//cc:defs.bzl",
-    "cc_library"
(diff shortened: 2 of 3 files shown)


.bazelrc
```diff
@@ -1,4 +1,3 @@
 # Force the tests to not run in sandboxed mode, since it can introduce
 # errors and flakes.
```
tinder-maxwellelliott, line 2:
```suggestion
```
Comment is no longer needed

BUILD
```diff
@@ -1,31 +1,26 @@
...
+    ]),
     hdrs = ["Sources/CYaml/include/yaml.h"],
+    # Requires because of https://github.com/bazelbuild/bazel/pull/10143 otherwise host transition builds fail
+    copts = ["-fPIC"],
     includes = ["Sources/CYaml/include"],
-    visibility = [
-        "//Tests:__subpackages__",
-    ],
+    linkstatic = True,
+    tags = ["swift_module"],
```
tinder-maxwellelliott, line 15: I would never know to do this :facepalm:

tinder-maxwellelliott (approved):

jpsim:
Thanks Keith.

PR #313, opened 2021-03-24
Simplify bazel CI configuration

keith:
This reduces the need to pass through some custom environment, and makes it work locally on Linux without copying this command. To do this I had to update the unit test target name because having a target named the same thing as a directory in your `data` results in a conflict where the `Tests/YamsTests` was actually a binary instead of a directory.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -42,7 +42,7 @@ jobs:
         shell: bash
       - name: Yams tests
         if: always()
-        run: bazelisk test //Tests:YamsTests
+        run: bazelisk test //Tests:UnitTests
         shell: bash
   Linux:
     strategy:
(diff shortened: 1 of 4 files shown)


.github/workflows/bazel.yml
```diff
@@ -62,9 +62,7 @@ jobs:
         shell: bash
       - name: Yams tests
         if: always()
-        run: |
-             build_workspace_directory=$(bazelisk info workspace)
-             CC=clang bazelisk test //Tests:YamsTests --action_env=PATH --test_env=BUILD_WORKSPACE_DIRECTORY=$build_workspace_directory
```
keith, line 67: I might need to maintain the `PATH` thing here (dependent on github actions), but if we do I think we can move it to the .bazelrc

jpsim: Looks like it’s still needed.

keith: I think there's a different issue with the fixture discovery change I made, not sure why it works for me locally, I'll have to debug some more

Tests/BUILD
```diff
@@ -13,17 +13,17 @@ genrule(
...
     srcs = ["LinuxMain.swift"],
     outs = ["main.swift"],
-    cmd = "cp $< $@"
+    cmd = "cp $< $@",
 )
 
 swift_test(
-    name = "YamsTests",
-    module_name = "YamsTests",
+    name = "UnitTests",
```
jpsim, line 20: Can we keep this named YamsTests?

keith: Checkout the updated description

jpsim: Ah yes that clears it up.

jpsim (approved):
This doesn’t risk breaking external users, does it?

keith:
It shouldn't!

Tests/BUILD
keith: found https://github.com/bazelbuild/bazel/issues/12312 for future reference

PR #315, opened 2021-04-05
Swap `sortKeys` bool for `keySort` function

kevboh:
(This is a breaking change, but I can make it non-breaking if preferred.)

This changes the `sortKeys` option to be a more flexible `keySort` function. The function defaults to `nil`—do not sort—but can be anything that sorts `Node`s. I ran into an issue where I wanted a custom sort on a mapping.


jpsim:
Thanks for the PR! Could you please rebase your commits on top of `main`? That should fix the failing CMake CI jobs.

kevboh:
@jpsim this has been rebased. Thanks for taking a look!

kevboh:
@jpsim let me know if anything else is required here.

kevboh:
Hi, is this still mergeable? If so I'd love to merge it!

lynchsft:
Hello! I'm sorry that the communication failed in 2021. The project is under more active maintenance now. 

I'm not a fan of the source-breaking change of turning `sortKeys` into `keySort`. 

 I _do_ however appreciate the ability to custom-sort the keys. 
 
 This is what I propose: 
 
 Introduce a `KeySort` enum type with:
 * a `.natural` case which performs the natural `<` sort
 * a `.custom((Key, Key)->Bool)` case for custom sorting. 
 * a `.none` case for non-sorting.
 
 If `KeySort` conforms to `ExpressibleAsBooleanLitteral` then `true` can map to `.natural` and `false` can map to `.none`. 
 
 With this arrangement we can gain custom sorting without breaking backward compatibility with uses of the existing sort behavior. 
 
 Thoughts?

kevboh:
I don't write Swift professionally anymore or use this project, so I defer to you!

PR #318, opened 2021-04-19
WIP: Run all tests on Windows

jpsim:
Some of these used to crash. Will try to integrate https://github.com/swift-server/swift-backtrace/pull/42 to troubleshoot failures if they still occur.


jpsim:
https://bugs.swift.org/browse/SR-14513

PR #319, opened 2021-05-12
Use Swift 5.4 release toolchain on Windows

HeMet:
I think it should fix https://bugs.swift.org/browse/SR-13819 I tried it myself and all seems work https://github.com/HeMet/Yams/actions/runs/836795843

### DIFF
--- a/.github/workflows/swiftpm.yml
+++ b/.github/workflows/swiftpm.yml
@@ -63,9 +63,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: seanmiddleditch/gha-setup-vsdevenv@master
-      - name: Install swift-DEVELOPMENT-SNAPSHOT-2021-02-18-a
+      - name: Install swift-5.4
         run: |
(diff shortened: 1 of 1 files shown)


jpsim:
Do all tests pass if you re-add the ones we disabled for Windows? Search for `TODO: Fix these tests on Windows` to see all the disabled tests.

HeMet:
> Do all tests pass if you re-add the ones we disabled for Windows? Search for `TODO: Fix these tests on Windows` to see all the disabled tests.

No, these tests fails without any backtrace. The program just terminates silently.

In `testFloat` `sexagesimal` is causing crash. In `testInt` all cases are problematic. Since these tests testing basic stuff maybe this functionality is the source of all other failures.

jpsim:
Right. I tried recently and filed https://bugs.swift.org/browse/SR-14513 to track the Windows crashes.

Nonetheless, it’ll be good to use an official release on CI rather than a snapshot.

HeMet:
While processing sexagesimal number crash occurs here: https://github.com/jpsim/Yams/blob/main/Sources/Yams/Constructor.swift#L530 `String.sexagesimal` doesn't get called. It looks like a compiler bug.

PR #320, opened 2021-05-15
Changed dump function parameter requirements

ABridoux:
## Summary
The `dump` functions now require a `NodeRepresentable` value to be passed rather than `Any`. 

### Context
- This makes the caller responsible for ensuring the provided value can be represented by a node, while default implementations for common types is provided.
- Also fixed some typos

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.0.7
+
+##### Breaking
+- The `dump` functions now requires `NodeRepresentable` values rather than `Any?` (or any `Sequence`) [#299]
+
 ## 4.0.6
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.0.7
```
jpsim, line 1: We won’t be making a patch release with this change because it is breaking.

jpsim:
```suggestion
## Main
```

lynchsft:
Hey @ABridoux this change still seems relevant. If you're interested in merging the PR, please rebase/merge and add tests covering the new behavior. :)

PR #326, opened 2021-06-22
Fix typos in documentation comments

mattt:
I was recently digging into Yams internals and noticed a couple typos in some documentation comments.


jpsim (approved):
Thanks

PR #332, opened 2021-11-13
CI: Update and simplify Windows setup

compnerd:
Update the Windows CI to use 5.4.3 instead of 5.4.0.  This update enables switching to gha-setup-swift which simplifies the CI setup for Windows.


jpsim:
Thanks, the Big Sur failures are unrelated and I’ll fix them separately.

PR #335, opened 2021-11-21
CI: enable multi-version builds on Windows

compnerd:
With the use of compnerd/gha-setup-swift, it is relatively easy to enable a
matrix build on Windows, enabling building and testing against different Swift
verisons.  Use this to enable builds for 5.5 and 5.6-dev.


jpsim:
Thanks!

PR #336, opened 2021-11-30
Tests: enable the full test suite on Windows with newer Swift

compnerd:
The underlying compiler issue that was prenventing the test suite from complete
execution on Windows has been resolved.  Restrict the full test suite to newer
versions of Swift.  While technically the 11/10 snapshot should contain the fix,
another issue prevents its usage.  The 11/20 snapshot and newer should support
the full test suite on windows as well.

The Windows support should now be at complete parity with the other platforms.

### DIFF
--- a/.github/workflows/swiftpm.yml
+++ b/.github/workflows/swiftpm.yml
@@ -84,7 +84,7 @@ jobs:
             tag: 5.5.1-RELEASE
 
           - branch: development
-            tag: DEVELOPMENT-SNAPSHOT-2021-06-12-a
+            tag: DEVELOPMENT-SNAPSHOT-2021-11-20-a
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


compnerd:
@jpsim - I think that this is likely the last of the windows changes - at this point, if you ignore the backwards compatibility checks, the only change for windows is to import `ucrt` rather than `Glibc` or `Darwin`.

jpsim:
Big Windows milestone, kudos for helping drive that! Merging shortly.

compnerd:
Thanks!  Happy to see this finally get resolved :)

jpsim:
I'm seeing a new issue when CI is running on main after merging this.

> C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swiftc.exe -print-target-info
TSCUtility/Triple.swift:154: Fatal error: could not determine host triple: malformed
error: unable to invoke subcommand: C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin\swift-build.exe ()

Any ideas what could be causing this?

compnerd:
No, but that is something that has shown up recently.  It's a flake, rerunning a couple of times makes it go away.  I'm hoping to try to see if I can reproduce it locally and understand what's the cause.

PR #343, opened 2022-01-30
Yams: make `CYaml` private to Yams

compnerd:
This changes how `CYaml` is built and used within Yams.  As there are no
interfaces from `CYaml` being directly exposed, we can statically link
the library.  However, this still would cause a problem as the `CYaml`
import would be serialized into the module requiring that the module is
available when building anything which consumes `Yams`.  To avoid that,
change the `import CYaml` instances to `@_implementationOnly`.  This
allows for `CYaml` to not be required at the consumer site.

While in the area of the CMake build system, clean up some of the build
system to use language specific options.  This is required to avoid the
de-duplication of the options across languages and enables linking
`CYaml` statically into the library.  Note that on Windows static
linking of *Swift* libraries is not yet properly supported and this does
not enable that nor use that - it is statically linking *C* content.

This now allows building Yams dynamically as a single library target.
On Windows, this is a negligible space savings of ~16KiB.

### DIFF
--- a/Sources/CYaml/CMakeLists.txt
+++ b/Sources/CYaml/CMakeLists.txt
@@ -1,27 +1,17 @@
 
-add_library(CYaml
+add_library(CYaml STATIC
   src/api.c
   src/emitter.c
   src/parser.c
   src/reader.c
   src/scanner.c
   src/writer.c)
-if(NOT BUILD_SHARED_LIBS)
-  target_compile_definitions(CYaml PRIVATE
-    YAML_DECLARE_STATIC)
-endif()
-target_compile_definitions(CYaml PRIVATE
-  YAML_DECLARE_EXPORT)
+target_compile_definitions(CYaml PUBLIC
+  $<$<COMPILE_LANGUAGE:C>:YAML_DECLARE_STATIC>)
 target_include_directories(CYaml PUBLIC
-  include)
+  $<$<COMPILE_LANGUAGE:C>:${CMAKE_CURRENT_SOURCE_DIR}/include>)
+target_compile_options(CYaml PUBLIC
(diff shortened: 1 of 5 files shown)


compnerd:
@jpsim this is likely something which is going to require some uglish workarounds for older versions of Swift.  What do you think is the best way forward with such a change?

jpsim:
> this is likely something which is going to require some uglish workarounds for older versions of Swift

More specifically, looks like this would bump the minimum version of Swift supported when building Yams to Swift 5.4, which was released [9 months ago](https://swift.org/blog/swift-5-4-released/). That seems like an acceptable amount of time to support older Swift versions to me.

Would you like to handle the removal of Swift < 5.4 CI jobs & code, or shall I?

jpsim:
Started working on bumping the minimum supported Swift version here: https://github.com/jpsim/Yams/pull/347

jpsim:
> I can take a stab at the codepaths for the older releases (if any) once the CI support is there.

Not sure what you mean, I don't think there will be code changes needed for Swift 5.4/5.5.

compnerd:
> Not sure what you mean, I don't think there will be code changes needed for Swift 5.4/5.5.

Ah right; there is only a single location (Tests/YamsTest/TopLevelDecoderTests.swift) that has a check for something other than 5.6.

jpsim:
@compnerd please rebase to get the changes from https://github.com/jpsim/Yams/pull/347 to require Swift 5.4.

compnerd:
That would be great - hopefully then I should be able to get this integrated into the swift-driver build.

jpsim:
Release tagged: https://github.com/jpsim/Yams/releases/tag/5.0.0

PR #354, opened 2022-06-05
Add DocC plugin and integrate with Swift Package Index

jpsim:
Following the instructions here: https://blog.swiftpackageindex.com/posts/auto-generating-auto-hosting-and-auto-updating-docc-documentation

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,0 +1,4 @@
+version: 1
+builder:
+  configs:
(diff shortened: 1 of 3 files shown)


finestructure:
FYI, we’ve removed the requirement to add the docc plugin yesterday but it’s not reflected in our docs yet!

PR #359, opened 2023-01-03
Remove rules_cc dependency

keith:
The previous plan to move the cc rules into this repo have been stalled. Having this load requires downstream users to include that repo that is just a shim to the main repo in the cc_library case.


jpsim:
Thanks!

PR #361, opened 2023-01-06
Update `actions/checkout` to v3

jpsim:


jpsim:
Some CI jobs are failing, I believe for unrelated reasons, so I won't update them here so they don't run.

PR #374, opened 2023-01-09
Add a way of disabling exponential notation when encoding doubles

### LINKED ISSUE #279: Doubles end up in exponential notation
I have the following object:

```swift
struct Example: Codable {
    var temperature: Double
    var latitude: Double
    var longitude: Double
}
```

And an example textual representation:

```swift
let example = """
temperature: 3.08
latitude: 40.7580
longitude: -73.9855
"""
```

I can decode it:

```swift
let decoder = YAMLDecoder()
let decodedExample = try decoder.decode(Example.self, from: example)
print(decodedExample)

// Example(temperature: 3.08, latitude: 40.758, longitude: -73.9855)
```

But when I try to encode it again the doubles all end up in exponential notation:

```swift
let encoder = YAMLEncoder()
let encodedExample = try encoder.encode(decodedExample)
print(encodedExample)

// temperature: 3.08e+0
// latitude: 4.0758e+1
// longitude: -7.39855e+1
```

How can I encode the doubles without the exponential notation?

zeionara:
Closes #279

In current version all doubles are decoded in a scientific notation:

```swift
import Yams

print(try Yams.dump(object: 6.85)))
```
produces:
```swift
6.85e+0
```

The proposed change is to add a static option to `Emitter.Options` class, so it can be used as follows:

```swift
import Yams

Emitter.Options.doubleFormatStyle = .decimal
print(try Yams.dump(object: 6.85)))
```

which produces the expected result:
```swift
 6.85
```

`YAMLEncoder` object is also supported, as well as numbers with zero fraction part, which are not saved as integers. For example, the following code:

```swift
import Yams

Emitter.Options.doubleFormatStyle = .decimal
print(try YAMLEncoder().encode(6.0)))
```

generates the expected result:

```swift
6.0
```


jpsim:
Very nice, thanks for the PR! I'll take a look shortly.

Sources/Yams/Emitter.swift
```diff
@@ -254,6 +259,14 @@ public final class Emitter {
 
         /// Set the style for mappings (dictionaries)
         public var mappingStyle: Node.Mapping.Style = .any
+
+        /// Set the style for formatting doubles
+        public static var doubleFormatStyle: NumberFormatStyle = .scientific
```
lynchsft, line 264: A `static var` is not compatible with Swift 6 language mode. If we refactor this as an instance variable we can revitalize this PR.

lynchsft (changes requested):
A static var is not compatible with Swift 6 language mode. If we refactor this as an instance variable we can revitalize this PR.

wojciech-kulik:
Any chance to revive this PR?

lynchsft (changes requested):
Please see comment dated Feb 24.

wojciech-kulik:
Oh, my mistake, I looked at the previous comment dated Jan 2013 and somehow assumed the next one from Feb was also from 2023 :D.

PR #377, opened 2023-01-13
Migrate to bzlmod

keith:
This switches over to using bzlmod for local testing, as well as creates a MODULE.bazel file that we can push to the central registry so other projects can depend on yams directly.


MODULE.bazel
```diff
@@ -0,0 +1,8 @@
+module(
+    name = "yams",
```
keith, line 2: unfortunately bzlmod is prescriptive about this and you cannot have capital letters

jpsim:
> creates a MODULE.bazel file that we can push to the central registry

Do I need to do something to do this?

MODULE.bazel
jpsim: it's fine

jpsim:
Thanks! Certainly makes this case simpler.

keith:
idk how much you want to sign up for for that, im fine to do submit it if you'd like, there is also automation we can configure with a github app for auto submitting updates

PR #380, opened 2023-01-13
Add bazel BCR configuration files

jpsim:
Should allow https://github.com/apps/publish-to-bcr to push releases to BCR automatically.

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,6 +1,24 @@
+shell_commands: &shell_commands
+- "echo --- Downloading and extracting Swift $SWIFT_VERSION to $SWIFT_HOME"
+- "mkdir $SWIFT_HOME"
+- "curl https://download.swift.org/swift-${SWIFT_VERSION}-release/ubuntu2004/swift-${SWIFT_VERSION}-RELEASE/swift-${SWIFT_VERSION}-RELEASE-ubuntu20.04.tar.gz | tar xvz --strip-components=1 -C $SWIFT_HOME"
+
 tasks:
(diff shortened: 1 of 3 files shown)


.bcr/source.template.json
```diff
@@ -0,0 +1,5 @@
+{
+    "url": "https://github.com/jpsim/Yams/archive/refs/tags/{TAG}.tar.gz",
+    "integrity": "",
+    "strip_prefix": ""
```
keith, line 4:
```suggestion
    "strip_prefix": "Yams-{TAG}"
```

.bcr/presubmit.yml
```diff
@@ -0,0 +1,9 @@
+tasks:
+  verify_targets:
+    name: Verify targets
+    platform: macos
+    build_targets:
+      - '@yams//:Yams'
+      - '@yams//Tests/...'
```
keith, line 7: can remove this since testing will build everything as long as there aren't build only targets in there, nbd either way

keith (approved):
looks good but you might want to test on linux too? https://github.com/keith/rules_multirun/blob/main/.bcr/presubmit.yml

jpsim:
> looks good but you might want to test on linux too?

Is Swift installed on BCR's Linux CI?

keith:
ah looks like you need this if so https://github.com/bazelbuild/bazel-central-registry/blob/2c81a61e94b31f3d5a8549d30a45f6de451ebf83/modules/rules_swift/1.5.1/presubmit.yml#L1-L4

keith (approved):

PR #383, opened 2023-01-17
Remove unused `buffer` property from `Parser.swift`

jpsim:
This was written to but never read.


jpsim:
Ah, turns out we need this to strongly retain the buffer. 🤦

PR #386, opened 2023-01-28
build: use static linking for `CYaml`

compnerd:
When building with SPM, the library will be built for static linking rather than dynamic linking.  Use the `YAML_DECLARE_STATIC` to control this behaviour.  Unlike CMake, we do not have fine-grained control over the artifact linkage.


jpsim (approved):
Do you need a release with this change?

compnerd:
Yes please!  That will allow the bump to the version in the dependents.

PR #388, opened 2023-02-03
Remove conditional docc-plugin dependency

### LINKED ISSUE #387: Documentation generation fails in Swift Package Index
I noticed that documentation generation fails in SPI at the moment. The reason is the following section in the package manifest:

https://github.com/jpsim/Yams/blob/4889c132978bc6ad3e80f680011ec3dd4fead90c/Package.swift#L4-L9

We're generating doc with Swift 5.7 which means the plugin is not available to us.

We have a process in place to scan for the docc plugin in the manifest and attach it ourselves if we can't find it. However, that process is pretty simple and basically just greps for the plugin url in the manifest - and in case of this manifest we find one and think the plugin is already available.

The easiest fix would be to remove this bit from the manifest if that's an option. Would you accept a PR to do so? Or do you need the conditional plugin dependency for other reasons?

finestructure:
Fixes #387


jpsim:
Can you please remove Package.resolved?

finestructure:
Ah of course!

jpsim:
Thanks!

PR #395, opened 2023-06-04
Allow decoding from a Node

rnapier:
This permits decoding the entire document into a Node, and then extracting Decodable types from parts of the document without creating a structure for the entire thing.


jpsim (approved):
Looks great, thanks for the contribution!

PR #398, opened 2023-07-18
build: pass `YAML_DEFINE_STATIC` to the CInterop

compnerd:
While we would now build CYaml statically, we would fail to indicate to clang the library was meant to be used statically and not dynamically. This cleans up some linker warnings when building SPM.


jpsim (approved):
Thanks!

Issue #405, opened 2024-02-07
Strings with newlines should be encoded as multiline literals


tejassharma96:
This code:
```
let testYaml = #"""
script:
  - |
    if [[ some test ]]; then
      echo "Hello world!"
    fi
"""#
let loaded = try Yams.load(yaml: testYaml) as? [String: Any]
print(loaded as Any)
```
prints
```
Optional(["script": ["if [[ some test ]]; then\n  echo \"Hello world!\"\nfi"]])
```

But when I try to encode something with newlines:
```
let dump = try Yams.dump(object: loaded)
print(dump)
```
prints
```
script:
- "if [[ some test ]]; then\n  echo \"Hello world!\"\nfi"
```
instead of using a multiline literal, it instead encodes them as `\n`. This is confusing, since Yams handles parsing these fine, but is unable to handle encoding them. If you'd like to maintain the current behaviour, I'd suggested adding something like this to `Emitter.Options`:
```swift
enum NewLineHandlingStrategy {
    /// Encode newlines in strings as \n
    case escapeNewlines
    /// Encode newlines in strings with a literal block, ie |
    case literal
    /// Encode newlines in strings with a foided block, ie >
    case folded
}
```

[Here](https://alisoftware.github.io/yaml/2021/08/19/yaml-part2-strings/)'s a reference on these

Maxatma:
```swift
/// Create `Emitter.Options` with the specified values.
    ///
    /// - parameter canonical:     Set if the output should be in the "canonical" format described in the YAML
    ///                            specification.
    /// - parameter indent:        Set the indentation value.
    /// - parameter width:         Set the preferred line width. -1 means unlimited.
    /// - parameter allowUnicode:  Set if unescaped non-ASCII characters are allowed.
    /// - parameter lineBreak:     Set the preferred line break.
    /// - parameter explicitStart: Explicit document start `---`.
    /// - parameter explicitEnd:   Explicit document end `...`.
    /// - parameter version:       The `%YAML` directive value or nil.
    /// - parameter sortKeys:      Set if emitter should sort keys in lexicographic order.
    /// - parameter sequenceStyle: Set the style for sequences (arrays / lists)
    /// - parameter mappingStyle:  Set the style for mappings (dictionaries)
    public init(canonical: Bool = false, indent: Int = 0, width: Int = 0, allowUnicode: Bool = false,
                lineBreak: Emitter.LineBreak = .ln, version: (major: Int, minor: Int)? = nil,
[...]

PR #406, opened 2024-02-07
Add an option for encoding scalars with newlines to Emitter.Options

tejassharma96:
Solves #405 by adding a new option for encoding scalars which contain newlines

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ##### Enhancements
 
-* None.
+* Allow specifying a `newLineScalarStyle` for encoding string scalars with newlines when using `YAMLEncoder`
+  [Tejas Sharma](https://github.com/tejassharma96)
+  [#405](https://github.com/jpsim/Yams/issues/405)
 
 ##### Bug Fixes
 
--- a/Sources/Yams/Emitter.swift
+++ b/Sources/Yams/Emitter.swift
@@ -41,7 +41,8 @@ public func dump<Objects>(
     version: (major: Int, minor: Int)? = nil,
     sortKeys: Bool = false,
     sequenceStyle: Node.Sequence.Style = .any,
-    mappingStyle: Node.Mapping.Style = .any) throws -> String
+    mappingStyle: Node.Mapping.Style = .any,
+    newLineScalarStyle: Node.Scalar.Style = .any) throws -> String
     where Objects: Sequence {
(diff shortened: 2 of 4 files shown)


acecilia:
Thanks! Looking forward to this. @jpsim wdyt? 🙏

danibachar:
+1 for this too

acecilia (approved):

Yams.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name                      = 'Yams'
-  s.version                   = '5.0.6'
+  s.version                   = '5.1.0'
```
danibachar, line 3: Version should be updated to 5.2.0

jpsim: Please don’t touch this file. I’ll bump the version when the time comes to cut a release.

jpsim (approved):
This looks good! Could you please add a change log entry and rebase? Thanks for your contribution.

tejassharma96:
ok rebased and updated the changelog 🙏

tejassharma96:
older swift versions do not understand `any Protocol` and the tests are failing because of that:
```
Encoder.swift:229:45: protocol 'StringProtocol' can only be used as a generic constraint because it has Self or associated type requirements
```

updated to add a `#if swift(>=5.6)`, which I _believe_ is when support for that was added, and for prior versions just try to cast to `String` instead of `any StringProtocol`

jpsim:
@tejassharma96 just making sure you see the CI failures here.

tejassharma96:
yes, sorry - somewhat distracted. I don't have an environment where I can install old versions of xcode since I'm on sonoma :/ how would you suggest I debug?

tejassharma96:
Ah actually - based on [this article](https://www.donnywals.com/what-is-the-any-keyword-in-swift/), the check should be for 5.7 instead of 5.6:
>  in Swift 5.7 the requirements around protocols with a Self requirement have been relaxed

Updated the check, could you see if CI succeeds now please?

tejassharma96:
hmmm, one failure still but looks unrelated to this change:
```
Failing tests:
	YamsTests:
		PerformanceTests.testUsingComposeWithUTF8()
		PerformanceTests.testUsingComposeWithUTF8()
```

will see if I can repro locally

jpsim:
CI is all ✅ 

I’ll merge shortly.

acecilia:
Thank you both 🙌

Issue #337, opened 2021-12-19
Issue with sexagesimal support


kean:
Hi, 

I've encountered an issue because of the sexagesimal support in Yams. When you work with scalars, you can expect the `Int` initializer to return `nil` in the following scenarios:

```swift
Int("a") // return nil
Int("a:b") // returns nil
```

But `YAMLDecoder` has support for "sexagesimal" numbers (which was dropped from the YAML spec in 2009). If you use `singleValueContainer` and try to decode `Int` using `YAMLDecoder`, like this:

```swift
let container = try decoder.singleValueContainer()
let value = try? container.decode(Int.self)
```

It hits the [`scalar.contains(":")`](https://github.com/jpsim/Yams/blob/9ff1cc9327586db4e0c8f46f064b6a82ec1566fa/Sources/Yams/Constructor.swift#L302) condition and at this point, the parser is convinced it is a number and never bails out, even if it's not a proper sexagesimal number. So the `Int.self` decoding produces the following result:

```swift
decodeIntFrom("a") // returns nil
decodeIntFrom("a:b") // returns `0`
decodeIntFrom("https://example.com/webhook") // returns `0`
```

It breaks certain assumptions that are true for `JSONDecoder`. I'm not sure if I did a great job describing the issue, but I suggest revising support of pre YAML 1.2 features.

kean:
I found the following note on [wiki](https://en.wikipedia.org/wiki/Sexagesimal#YAML):

````
In version 1.1[17] of the YAML data storage format, sexagesimals are supported for plain scalars, and formally specified both for integers[18] and floating point numbers.[19] This has led to confusion, as e.g. some MAC addresses would be recognised as sexagesimals and loaded as integers, where others were not and loaded as strings. In YAML 1.2 support for sexagesimals was dropped.[20]
```

That's basically exactly the issue I'm encountering.

PR #409, opened 2024-03-15
Fix parsing of unquoted URLs

czechboy0:
## Summary

I believe this fixes https://github.com/jpsim/Yams/issues/337, which I also just hit.

The problem is that trying to parse an int _succeeded_ for the unquoted string `http://example.com` (parsed as the number `0`), which caused issues when using an "AnyCodable" approach.

This PR makes the parsing stricter, so that a URL like above does not successfully parse as an int, and allows "AnyCodable" wrappers to only successfully parse the value as a string.

## Test Plan

Added a unit test both for the positive and negative cases.


jpsim (approved):
Looks great, thank you so much for the fix!

PR #413, opened 2024-03-24
Fix a typo in README.md

1024jp:
This is just a fix of a typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ cmake --build /path/to/build
 ```
 
 To build for Apple platforms (macOS, iOS, tvOS, watchOS), there is no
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,7 +28,7 @@ cmake --build /path/to/build
 ```
 
 To build for Apple platforms (macOS, iOS, tvOS, watchOS), there is no
-need to spearately build Foundation because it is included as part of
+need to sparately build Foundation because it is included as part of
```
jpsim, line 31:
```suggestion
need to separately build Foundation because it is included as part of
```

jpsim (approved):
Thanks!

PR #416, opened 2024-05-06
Fail on duplicate keys in a mapping

### LINKED ISSUE #415: Yams should fail if there are duplicate keys in a mapping
I recently encountered a bug when using Yams to read and write some yaml, and realized that the cause was a duplicate key in my yaml file that was being handled silently. 

Per the [yaml spec](https://yaml.org/spec/1.2-old/spec.html#id2764044):
> The content of a mapping node is an unordered set of key: value node pairs, with the restriction that each of the keys is [unique](https://yaml.org/spec/1.2-old/spec.html#equality//). YAML places no further restrictions on the nodes. In particular, keys may be arbitrary nodes, the same node may be used as the value of several key: value pairs, and a mapping could even contain itself as a key or a value (directly or indirectly).

Basically, duplicate keys are not allowed. Yams should recognize when there is a duplicate key in a mapping and throw an error rather than having undefined behaviour (I think it just picks the version that is lowest in the file).

As an example:
```
import Foundation
import Yams

struct MyCodable: Codable {
    let thing: String
}

let yaml = """
key:
  thing: hello
key:
  thing: goodbye
"""

let data = Data(yaml.utf8)

let decoder = YAMLDecoder()
let test = try decoder.decode([String: MyCodable].self, from: data)
print(test)
```

prints `["key": __lldb_expr_7.MyCodable(thing: "goodbye")]` - it doesn't throw any errors and selects the key that is lower in the text.

tejassharma96:
This fixes #415 but it is a slightly "naive" implementation. It's possible that the "correct" place to throw an error from would be the C code rather than the swift code, but I do not have the knowledge to make those changes 😓 

Please let me know what you think!


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+
+* None.
+
+##### Enhancements
+
+* None.
+
+##### Bug Fixes
+
+* Yams will now correctly error when it tries to decode a mapping with duplicate keys.
```
tejassharma96, line 13: This is technically a bug fix but it may be breaking if people have malformed yaml they are trying to decode with this

jpsim:
Thanks for the PR!

This is a welcome change, and doing this at the Swift layer is fine. Based on some GitHub discussions, it does look like the C API can detect duplicate keys. There's also this issue with relevant discussion: https://github.com/yaml/pyyaml/issues/41

However, I think it's important that if we start to error in these cases, that 1) it be recoverable (e.g. we don't fail the whole decoding) and 2) the error is reported at the correct location (mark), right now 1:1 is being hardcoded as the location.

If you rebase or merge main you should have CI in a healthier state to run again.

tejassharma96:
hmmm ok, I'll need a bit of support on both of those due to my lack of familiarity here:  

1. Recoverable - what would you suggest here? I personally would actually expect the entire decoding to fail if we have invalid YAML. I'm curious what you're thinking regarding how we can both make the error clear and visible to users but also not fail the whole decoding.
2. Correct location - yeah, I was wondering what to do here, couple questions:
    - Say there are 3 instances of a specific key, at which one of those keys should we error? Can we even decide that? I suppose we could use `key.mark` and compare line/column numbers to choose either first/last in the data? 
    - What if there are multiple instances of duplicated keys? If keys a, b and c are duplicated, currently my code would basically just error on the first one, but the error would say `but found multiple instances of: ["a", "b", "c"]`. I think there's a balance to be struck here on being specific about the point of error but also giving the consumer actionable feedback on all the issues. I suppose this could potentially be an opportunity to create a new case on `YamlError`, something like `.duplicatedKeys(keys: [Node])` that could display all the relevant information from the nodes (ie by fetching `.mark` for all duplicated nodes and creating some custom error message that gives immediate feedback about all duplicated keys)?

tejassharma96:
Updated the error to notify of all duplicated keys and their locations, haven't updated it to be recoverable bc personally I don't believe it _should_ recover - imo it should just throw immediately

Sources/Yams/Parser.swift
```diff
@@ -355,13 +355,32 @@ private extension Parser {
...
         let node = Node.mapping(.init(pairs, tag(firstEvent.mappingTag), event.mappingStyle, firstEvent.startMark))
         if let anchor = firstEvent.mappingAnchor {
             anchors[anchor] = node
         }
         return node
     }
 
+    private func checkDuplicates(mappingKeys: [Node]) throws {
+        var duplicates: [String: [Node]] = [:]
+        for key in mappingKeys {
```
tejassharma96, line 369: this for loop could just be
```
let duplicateKeys = Dictionary(grouping: keys, by: { $0.string! }).filter { $1.count > 1 }.keys
```
but I wanted to avoid the force unwrap. if we're confident that these keys will always be scalars (which I think they _should_ be but I'm not sure about that) then I can just switch to that simpler form

tejassharma96:
hey @jpsim, apologies for the tag but I'd love another review here when you have a minute 😅

Sources/Yams/Parser.swift
lynchsft: What about? 
```suggestion
        let duplicateKeys = Dictionary(grouping: keys) { 
                                                                $0.string ?? "[non-scalar-key]
                                                            }
                                                            .filter { $1.count > 1 }
```

But actually Node.Mapping has a .string value too via Node.string -> String.construct(from:)

And actually [Yaml.org](https://yaml.org/spec/1.2.2/#742-flow-mappings) 
Header: "Example 7.21 Single Pair Implicit Entries" seems to indicate that non-scalar keys are allowed?

So perhaps what would be most correct here here was to put the nodes in a Set and see if the set has fewer elements than the keys list. 

Or Yams could just document the behavioral deficiency. It's pretty weird to use a key that's not a scalar. 😆

tejassharma96: oh, interesting! didn't realize that having non-scalar keys was allowed. I like the idea of using a `Set`, but after giving implementing this a try I realized that the existing structure makes it a lot easier to print a helpful error message when this issue pops up. 

That said, given that `Node`s are `Hashable`, I can just use them as the dictionary keys. This works since the metadata isn't used in `hash(into:)`, only the tag and value ([eg for Node.Scalar](https://github.com/jpsim/Yams/blob/main/Sources/Yams/Node.Scalar.swift#L91)) (otherwise the same key in a different line would hash to a different value). Updated and the tests I wrote earlier (back in July, wow!) continue to succeed so I think this method works better. Thanks for the suggestion!

tejassharma96:
guessing the CI error is the same one in https://github.com/jpsim/Yams/pull/428 so I just copied over the fix (thanks @lynchsft!)

danibachar:
+1 to this change, would love to see it in the upstream branch

bradleymackey (approved):

lynchsft (approved):

tejassharma96:
Thanks @lynchsft!

PR #419, opened 2024-06-16
add support for riscv64

futurejones:
Add support for `riscv64` as host architecture.


futurejones:
@jpsim can you take a look at this please.

futurejones:
@jpsim just wondering if there is a problem with this PR?

jpsim:
Thanks for the PR, no there's no problem with it.

PR #420, opened 2024-06-29
Bazel: support rules_swift 2.x

luispadron:
rules_swift 2.x increases the `compatibility_level` so we need to declare we support up-to this new level.


jpsim:
Thanks! Can you please update the CI workflow file to include this file so CI runs when it’s modified? https://github.com/jpsim/Yams/blob/main/.github/workflows/bazel.yml

luispadron:
Updated! Would you also own `sourcekitten_com_github_drmohundro_SWXMLHash`? I can't find the repo with that file but we'll need to make a similar change there if it's using `rules_swift`

jpsim:
I think that’s defined in https://github.com/jpsim/SourceKitten

jpsim (approved):

PR #418, opened 2024-05-28
add os(visionOS) support

ruralharry:
As mentioned in the title

PR #422, opened 2024-07-09
GitHub Actions workflow fixes

tonyarnold:
Apologies for the notification spam while I was working on this issue. 

This PR proposes fixes for some of the workflow issues introduced with #418, and to update the GitHub Actions workflows where appropriate to use recent versions of Xcode, macOS and the Actions plugins.

This ended up being a wonderful confluence of problems to solve, resulting in the following changes:
- macOS 11 runners have been removed (GHA doesn't support them anymore).
- Swift versions prior to 5.7 on platforms other than macOS had to be removed, as the images were conflicting with the actions/checkout action (in reality, all node-based actions would fail on these older images).
- Where possible, base tasks like SwiftLint and CocoaPods verification happen on a macOS 14 Sonoma image running Xcode 15.4 (the latest at the time of authoring).
- CocoaPods needed to be upgraded for visionOS support.
- A few minor SwiftLint violations introduced in recent versions needed to be fixed or disabled.
- I disabled running SwiftLint from within the Xcode project when it detects that it is being built on CI. It seemed unnecessary given it's running in a dedicated workflow.
- Existing workflow jobs will now auto-cancel if a new set of jobs is started on the same branch.


tonyarnold:
😅 this is going to take some elbow grease.

tonyarnold:
Okay, I believe everything is now running on CI and passing out of the box. I'm more than happy to consolidate commits, or make further changes — just let me know!

PR #423, opened 2024-07-10
Use aspect hint for CYaml in Bazel

luispadron:
rules_swift 2.0 requires using the `aspect_hints` attribute to specify interop between `cc_library` and `swift_library`. This change adds the `aspect_hints` attribute to the `cc_library` rule for CYaml. It increases the minimum supported rules_swift version to 2.0

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,7 +5,7 @@ module(
 )
 
 bazel_dep(name = "apple_support", version = "1.15.1")
-bazel_dep(name = "rules_swift", version = "1.18.0", repo_name = "build_bazel_rules_swift", max_compatibility_level = 2)
+bazel_dep(name = "rules_swift", version = "2.0.0", repo_name = "build_bazel_rules_swift")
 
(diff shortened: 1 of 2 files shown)


brentleyjones (approved):

MODULE.bazel
```diff
@@ -5,7 +5,7 @@ module(
 )
 
 bazel_dep(name = "apple_support", version = "1.15.1")
-bazel_dep(name = "rules_swift", version = "1.18.0", repo_name = "build_bazel_rules_swift", max_compatibility_level = 2)
+bazel_dep(name = "rules_swift", version = "2.0.0", repo_name = "build_bazel_rules_swift")
```
brentleyjones, line 8: We should wait until 2.1.1 is approved, and use that, since it has a big bug fix in it.

luispadron:
Because this has a custom module map which is found via includes, we don't actually need to use aspect hints / increase rules_swift version

PR #424, opened 2024-07-10
Require Swift 5.7 or later

jpsim:
Matching other projects in the Swift compiler ecosystem. E.g. https://github.com/swiftlang/swift-driver/blob/70554a4dc01f7152a2d77597139957a1475138ff/Package.swift#L1

I'll bump the Yams version to 6.0 when releasing this.


tonyarnold:
I :heart: a good Swift version bump — look at all of the conditional code you were able to remove!

PR #431, opened 2024-10-17
Security fix in libyaml: Limit depth of nesting by default

hi-kumar:
Update libyaml:
- Bring over security fix in libYaml: 51843fe (PR 287)
```
Each nesting level increases the stack and the number of previous
starting events that the parser has to check.

The default maximum is 1000 and can be set via yaml_set_max_nest_level()
```
- While here, bring in couple more commits in the same code segments to stay up to date: `Handle closing flow sequence after explicit key` 588eabf (PR 295) & 840b65c (296)


jpsim (approved):
Thank you!

lynchsft (approved):

lynchsft (approved):

PR #434, opened 2024-12-01
build: fix amd64/AMD64 build

khng300:
The target name "amd64" is missing from the matching.


lynchsft (commented):
Hi! 

Thank you for your contribution. :) 
Can you clarify the meaning of this change? (I'm not very familiar with cmake.)

In what context does the code fail to build without this change? Can you add workflow files (or modify the existing ones) to demonstrate the problem and validate the fix? 

Thank you!

PR #428, opened 2024-09-20
Implement Anchor coding, Tag coding and redundancy auto-aliasing

lynchsft:
* Yams is able to encode and decode Anchors via YamlAnchorProviding, and
  YamlAnchorCoding. 
  [#125](https://github.com/jpsim/Yams/issues/125)
  
* Yams is able to encode and decode Tags via YamlTagProviding
  and YamlTagCoding.
  [#265](https://github.com/jpsim/Yams/issues/265)
  
* Yams is able to detect redundant structes and automaticaly
  alias them during encoding via RedundancyAliasingStrategy

PR #435, opened 2024-12-02
Reference coalescing

lynchsft:
Yams is able to coalesce references to objects decoded with YAML anchors.
 
The main use-case for dereferencing aliases in a YML document is when decoding into class types. If the yaml document is large and contains many references (perhaps it is a representation of a dense graph) then, decoding into structs will require the of large amounts of system memory to represent highly redundant (duplicated) data structures.

However, if the same document is decoded into class types and the decoding uses an `AliasDereferencingStrategy` such as `BasicAliasDereferencingStrategy` then the emitted value will have its class references coalesced. No duplicate objects will be initialized. In some scenarios this may significantly reduce the memory footprint of the decoded type.

This PR builds on concepts from an earlier PR: https://github.com/jpsim/Yams/pull/428


PeqNP:
Do any of these new strategies require a test to ensure they work as expected? Or do the current tests do a good job of covering what was added?

lynchsft:
> Do any of these new strategies require a test to ensure they work as expected? Or do the current tests do a good job of covering what was added?

@PeqNP  I believe the new reference coalescing strategies are covered by the tests. This PR is now _much_ easier to read after having merged #428 . Your review is appreciated! 🙏🏻

lynchsft:
Additional tests covering the AliasingStrategies have [been added here](https://github.com/jpsim/Yams/pull/438/files#diff-e12536c9813f9c1b09f70834af24f53398be6b56e6cf87db0bfa8a456f5ea62d)

PR #436, opened 2024-12-02
Use builtin `pow`

### LINKED ISSUE #429: Musl support
The combination of Yams and the Static Linux SDK allows the creation of a CLI in a single binary. To do this, the implementation needs to be changed from a direct dependency on Glibc to a dependency on both Glibc and Musl.

Here, import needs to be changed to support Musl.

```
#elseif canImport(Glibc)
import CoreFoundation
import Glibc
private let cpow: (_: Double, _: Double) -> Double = Glibc.pow
#elseif canImport(Musl)
import CoreFoundation
import Musl
private let cpow: (_: Double, _: Double) -> Double = Musl.pow
#endif
```

bradleymackey:
- Removes dependency on libc and the platform-specific `pow` function. This allows Linux builds using non-Glibc (like Musl) to build without error.
- `pow` should now be available in [Swift's standard library](https://github.com/swiftlang/swift/pull/23140/files) for all platforms and distributions.
- This workaround was originally added to support the Swift for Tensorflow (#281), but that project has now been archived so support is no longer required here.
- Resolves #429 
- Closes #430

### DIFF
--- a/Sources/Yams/Representer.swift
+++ b/Sources/Yams/Representer.swift
@@ -8,22 +8,6 @@
 
 import Foundation
 
-#if os(iOS) || os(macOS) || os(watchOS) || os(tvOS) || os(visionOS)
-import Darwin
-private let cpow: (_: Double, _: Double) -> Double = Darwin.pow
-#elseif os(Windows)
-import ucrt
-private let cpow: (_: Double, _: Double) -> Double = ucrt.pow
-#elseif canImport(Bionic)
-import CoreFoundation
-import Bionic
-private let cpow: (_: Double, _: Double) -> Double = Bionic.pow
-#else
-import CoreFoundation
-import Glibc
-private let cpow: (_: Double, _: Double) -> Double = Glibc.pow
-#endif
-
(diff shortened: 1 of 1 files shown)


lynchsft:
💪

lynchsft (approved):

PeqNP (approved):

fpseverino:
Having support for Musl would be really useful. Any updates on when it might be merged and released? @jpsim

SimplyDanny (approved):

lynchsft (approved):

bradleymackey:
I think it might be time fork so we can unblock SwiftLint's Linux build. If/when the changes are merged upstream here, we can revert to the primary repository. Thoughts @SimplyDanny?

bradleymackey:
Disregard my previous comment! Thanks @lynchsft, can we have a release cut as well? 😄

lynchsft:
Working on it this morning. More PR reviews and merges to follow. 
Much thanks to @jpsim for inviting me to collaborate. 😁

PR #439, opened 2025-02-21
Do not detect C++ toolchain

SimplyDanny:
This was necessary in the [previous Bazel release](https://github.com/bazelbuild/bazel-central-registry/pull/2404). So I guess this should be part of the template.

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -5,3 +5,5 @@ tasks:
     bazel: 7.x
     build_targets:
       - '@yams//:Yams'
+    build_flags:
+      - "--repo_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1"


keith (approved):

lynchsft:
Hey @SimplyDanny. Could you explain more about this change? The Bazel builds in the current workflow are passing. What problem / break does this address? Can you add/change the workflows to show how this change is needed/used?

SimplyDanny:
> Hey @SimplyDanny. Could you explain more about this change? The Bazel builds in the current workflow are passing. What problem / break does this address? Can you add/change the workflows to show how this change is needed/used?

This pre-submit workflow only runs in Bazel registry builds as a basic check. It highly depends on the infrastructure provided there. Projects submitting packages are free to customize the pipeline for whatever they think is necessary and reasonable to verify.

With the proposed build flag, the pipeline skips checking for a valid C++ toolchain. The same change was also made in the previous release (see link in the PR description). But it was only a one-time adaption without an update of the template.

PR #440, opened 2025-03-13
Don't decode explicit strings as bools, regardless of content

czechboy0:
## Summary

A scalar with the text `true`, for example, can be decoded as a Bool today - that's expected.

However, when using a YAMLDecoder, the scalars `!!str true`, `"true"`, and `'true'` also successfully decode as Bools - that's unexpected, especially with Decodable implementations often trying multiple different types to decode as, and use the first one that works.

The suggested change checks the scalar's tag property, and if it's explicitly marked as a string, doesn't allow creating a Bool value from it.

This fixes AnyCodable-style decoding of YAML values, which in turn affects correctness of round-tripping YAML through Yams and back into YAML.

## Test Plan

Added unit tests, verified they failed before this change and pass with the change.

### DIFF
--- a/Tests/YamsTests/NodeDecoderTests.swift
+++ b/Tests/YamsTests/NodeDecoderTests.swift
@@ -33,4 +33,64 @@ final class NodeDecoderTests: XCTestCase {
         XCTAssertEqual(desired.name, "My Name")
         XCTAssertEqual(desired.age, 123)
     }
+
+    func testDecodeBools() throws {
+        let yaml = """
+        ---
+        topLevel:
+          unquotedBool: true
+          explicitBool: !!bool true
+          explicitStringNotBool: !!str true
+          singleQuotedStringNotBool: 'true'
+          doubleQuotedStringNotBool: "true"
+        """
+
+        struct TopLevel: Decodable {
+            var unquotedBool: BoolOrString
+            var explicitBool: BoolOrString
+            var explicitStringNotBool: BoolOrString
+            var singleQuotedStringNotBool: BoolOrString
+            var doubleQuotedStringNotBool: BoolOrString
+        }
+
+        let node = try Yams.compose(yaml: yaml)!
+
+        let desiredNode = try XCTUnwrap(node["topLevel"])
+
+        let desired = try YAMLDecoder().decode(TopLevel.self, from: desiredNode)
+
+        XCTAssertEqual(desired.unquotedBool, true)
+        XCTAssertEqual(desired.explicitBool, true)
+        XCTAssertEqual(desired.explicitStringNotBool, "true")
+        XCTAssertEqual(desired.singleQuotedStringNotBool, "true")
+        XCTAssertEqual(desired.doubleQuotedStringNotBool, "true")
+    }
+}
+
+enum BoolOrString: Equatable {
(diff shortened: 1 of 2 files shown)


czechboy0:
The Windows CI fail doesn't seem to be caused by this PR, I don't think?

lynchsft:
Thank you for the PR! 

This change looks great! 😊 I'd like to request one small addition to the test you've added. See review comments. 🙂

I'm unsure what's causing the windows failure. It looks like VSCode isn't setup properly. Will spend some time Investigating.

Tests/YamsTests/NodeDecoderTests.swift
```diff
@@ -33,4 +33,64 @@ final class NodeDecoderTests: XCTestCase {
...
+        let desiredNode = try XCTUnwrap(node["topLevel"])
+
+        let desired = try YAMLDecoder().decode(TopLevel.self, from: desiredNode)
+
+        XCTAssertEqual(desired.unquotedBool, true)
+        XCTAssertEqual(desired.explicitBool, true)
+        XCTAssertEqual(desired.explicitStringNotBool, "true")
+        XCTAssertEqual(desired.singleQuotedStringNotBool, "true")
+        XCTAssertEqual(desired.doubleQuotedStringNotBool, "true")
+    }
```
lynchsft, line 67: This is a great test, thank you. 

I think your point/intention would be made even more clearly and strongly if we add a call to `_testRoundTrip` here. 
Particularly, I think it would be helpful to show that the decoded form has `XCTAssertNotEqual(desired.doubleQuotedStringNotBool, .bool(true))`

lynchsft (changes requested):
I very much approve, just wanting to make the test as explicit as possible.

Tests/YamsTests/NodeDecoderTests.swift
czechboy0: I tried `_testRoundTrip`, but the out YAML doesn't include the three different ways of encoding the string, so it failed.

I did update the equality assertions though - I agree, it's much clearer when not using ExpressibleBy... protocols: 3b15363

lynchsft (approved):

Tests/YamsTests/NodeDecoderTests.swift
lynchsft: I see your point. We'd have to introduce a funny type that conforms to YamlTagProviding and RawRepresentable and basically it's weird. 
I've approved. ✅😁

The windows failures are not due to this change. I re-ran the checks on another branch and they are failing the same way. I'm not sure what the solution is right now.

czechboy0:
Thanks @lynchsft, I'll let you merge when you see fit.

lynchsft:
@jpsim Do you understand the CI failure on windows. The matrix runs in parallel, I've seen at lease three different swift versions fail the same way so I don't think it's related to any swift version. 
If you recognize the issue can you suggest a course of action?

lynchsft:
@czechboy0 The windows failures are now resolved in the main branch. Please merge/rebase. :)

czechboy0:
> @czechboy0 The windows failures are now resolved in the main branch. Please merge/rebase. :)

Done! 🙏

PR #441, opened 2025-03-19
Enable library evolution to support Swift 6.1

pennig:
Xcode 16.3/Swift 6.1 now treat usages of `@_implementationOnly` as errors unless library evolution is enabled. 

Updates _BUILD_ and fixes an issue with unsupported scoped imports in module interfaces.

### DIFF
--- a/Sources/Yams/Emitter.swift
+++ b/Sources/Yams/Emitter.swift
@@ -7,7 +7,11 @@
 //
 
 #if SWIFT_PACKAGE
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks @pennig ! Based on the CI failures, do we need to bump the minimum required rules_swift for this?

pennig:
Ahh, yeah. Looks like we'll need >=2.2.0

Issue #442, opened 2025-03-24
Yaml Anchors Breaking as of 5.3.0


alteredtech:
I am working on a toy problem for work with the swift-openapi-generator. I have recently encountered an error where it no longer builds. After some investigation on this issue https://github.com/apple/swift-openapi-generator/issues/746 I was able to narrow it down to this PR https://github.com/jpsim/Yams/pull/428 for `5.3.0`. 

When I attempt to build on `5.3.0` I get this message

```
Error: /Users/alteredtech/Coding/swift-petstore-oas/Sources/openapi.yaml: error: Found neither a $ref nor a Response in .responses.200 for the **PUT** endpoint under `/pet`.

Response could not be decoded because:
Inconsistency encountered when parsing `Vendor Extension`: Found at least one vendor extension property that does not begin with the required 'x-' prefix. Invalid properties: [ yamlAnchor ]..
```

However, on `5.1.3` and `5.2.0` I do not and the build is successful.

My working commit example: https://github.com/Altered-Tech/swift-petstore-oas/tree/yaml-anchors
My breaking commit example: https://github.com/Altered-Tech/swift-petstore-oas/tree/yaml-anchors-fails

mattpolzin:
To expound, the problem OpenAPIKit has is that the structure it’s Codable types are being passed appears to have a meta `yamlAnchor` property (i.e. the property did not appear in the original YAML source) and OpenAPIKit expects to be able to assert that the input structure has no unexpected properties.

jpsim:
Thanks for the report! @lynchsft this seems related to your changes in https://github.com/jpsim/Yams/pull/428, could you please take a look?

lynchsft:
Has this issue been tested with Yams 5.3.1? It [contains anchor-related bug fixes.](https://github.com/jpsim/Yams/releases/tag/5.3.1)

lynchsft:
@mattpolzin Do we have contact with the OpenAPIKit maintainers? It sounds like they were pasting-on some anchor processing to a version of Yams prior to the introduction of first-party support. If that's truly the case, they should simply update their logic by deleting the third-party anchor handling.

mattpolzin:
That'd be me; I maintain OpenAPIKit. I try not to trust my own memory going back too far, but the word `yamlAnchor` does not appear in the OpenAPIKit codebase and it doesn't appear in the repo linked to in the OP above so I assumed that Yams was injecting that property somehow during decoding.

lynchsft:
[...]

PR #449, opened 2025-05-10
Meta keys hidden

lynchsft:
Addresses #442


lynchsft:
The exposure of meta-keys is shown in [this branch's new tests](https://github.com/jpsim/Yams/pull/448/files). The same tests pass with this PR's fix.

Issue #358, opened 2022-12-09
Yams without Foundation?


t089:
The state of Foundation on non Apple platforms continues to be less than ideal and brings the additional burden of increased binary size and memory footprint. Many libraries take effort the factor out their Foundation dependency in a separate module (eg. [karwa/swift-url](https://github.com/karwa/swift-url/tree/main/Sources/WebURLFoundationExtras), [apple/swift-nio](https://github.com/apple/swift-nio/tree/main/Sources/NIOFoundationCompat), discussion [swift-server/async-http-client](https://github.com/swift-server/async-http-client/issues/299), ...).

As an example, a simple "Hello World" Swift binary compiled with `swift build -c release --static-swift-stdlib` on `swift:5.7.1` weighs in at `7.6M`. Adding the `Yams` dependency increases the binary size to `50M`.

Of course, not relying on Foundation is not trivial, but would be good to at least map out what would need to be done to make it happen.

t089:
🤔 Bad timing, maybe there is something brewing at the fruit company.
https://twitter.com/ktosopl/status/1601145297077686272?s=46&t=C2w1wZhSOLAErfvayJ24KQ

jpsim:
> As an example, a simple "Hello World" Swift binary compiled with `swift build -c release --static-swift-stdlib` on `swift:5.7.1` weighs in at `7.6M`. Adding the `Yams` dependency increases the binary size to `50M`.

Yikes, that's a pretty heavy dependency.

Why do you need the statically compile the Swift stdlib if you're running it on what appears to be a Docker image with Swift installed?

On macOS, a simple `import Yams; print(try Yams.load(yaml: "a: 0")!)` CLI executable is 920 KB when built for release for arm64.

t089:
> Why do you need the statically compile the Swift stdlib if you're running it on what appears to be a Docker image with Swift installed?

Usually, we build on docker swift:5.7 but then deploy on a distroless image which has no (swift) runtime dependencies pre-installed to make the final image as small as possible.

jpsim:
ok that makes sense.

I hacked up a proof-of-concept to remove Foundation here: https://github.com/jpsim/Yams/compare/jp-no-foundation

It removes a ton of useful features from Yams, but it gets the following compiling without Foundation:

```swift
import Yams

let yaml = """
a: 0
b: 1.2
c: [1, 2, 3]
d:
  - a
  - b
  - c
"""

dump(try Yams.load(yaml: yaml)!)
```

[...]

PR #453, opened 2025-06-02
Depend only on Foundation Essentials

lynchsft:
Addresses #358

N.B. @t089

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,6 +3,7 @@ import PackageDescription
 
 let package = Package(
     name: "Yams",
+    platforms: [.iOS(.v13), .macOS(.v13), .tvOS(.v13), .watchOS(.v6)],
     products: [
         .library(name: "Yams", targets: ["Yams"])
     ],
--- a/Package@swift-6.swift
+++ b/Package@swift-6.swift
@@ -3,6 +3,7 @@ import PackageDescription
 
 let package = Package(
     name: "Yams",
+    platforms: [.iOS(.v13), .macOS(.v13), .tvOS(.v13), .watchOS(.v6)],
     products: [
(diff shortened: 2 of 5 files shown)


Sources/Yams/Parser.swift
```diff
@@ -396,6 +400,9 @@ private extension Parser {
 
     private func checkDuplicates(mappingKeys: [Node]) throws {
         let duplicates: [Node: [Node]] = Dictionary(grouping: mappingKeys) { $0 }.filter { $1.count > 1 }
+            .filter { (key, _) in
```
t089, line 403: How does this change relate to the goal of the PR?

lynchsft: It was mistakenly included from another experiment. Allow me a moment to clean up. :)

lynchsft:
@t089 This effort has lead to the necessity of replacing NumberFormatter, which I think is a really terrible decision for all the platforms that _do_ have NumberFormatter. 

I've engaged this thought process so far as an abstract exercise that the smallest footprint is the best footprint. But in the face of this challenge, I'll like to as a more practical question. What kind of platform would actually benefit from diminishing the dependency to merely FoundationEssentials? 

Microcontrollers? Kernel modules? What was on your mind when you initially opened the issue? What is on your mind now?

t089:
That is unfortunate indeed. I think the original observation is valid for any non-Darwin platform:

> As an example, a simple "Hello World" Swift binary compiled with swift build -c release --static-swift-stdlib on swift:5.7.1 weighs in at 7.6M. Adding the Yams dependency increases the binary size to 50M.

For a fundamental library such as yams it feels wrong to incur this binary size cost.

PR #460, opened 2025-08-12
feat: support DecodableWithConfiguration

nekrich:
### Summary

This PR introduces support for decoding types conforming to the new `DecodableWithConfiguration` protocol.

This enables more flexible and context-aware decoding by allowing strongly typed configuration to be passed during decoding — unlike `userInfo`.

### What's Changed

- Added support for decoding `DecodableWithConfiguration` objects.
- Refactored existing decoding paths to maintain a single `_Decoder` construction and Node parsing logic across calls to decode `Decodable` and `DecodableWithConfiguration`.
- New public API is the same as in the `JSONDecoder`.

### Why

Some types require external context (e.g., formatting rules, environment settings, or feature flags) when being decoded. `DecodableWithConfiguration` enables passing that context directly into the decoding process without affecting other parts of the codebase.

### Tests
Provided for both cases:
- Decoding with direct `DecodingConfiguration` instance.
- Decoding with `DecodingConfigurationProvider` constructor.

### Breaking changes
None. This change is purely additive and introduces no breaking changes to the existing API.


lynchsft:
Only one lint issue left to tic then we'll merge and release. 😀

PR #462, opened 2025-09-09
Use an aspect hint instead of the `swift_module` tag

brentleyjones:
This is needed when using rules_swift 2.x+, otherwise inputs aren’t properly tracked and RBE doesn’t work.


brentleyjones:
After this change Yams will only work with Swift 5.9 (maybe 5.8) with Bazel. Can the CI be adjusted for that @jpsim?

luispadron (approved):

PR #467, opened 2025-11-06
Modify CYaml header includes for Swift C++ compatibility

johnfairh:
Moving a large Swift-C++ project to Xcode 26 / macOS 15 / arm  I hit this error in Xcode about CYaml/yaml.h:

Screenshot 2025-11-06 at 12 28 35

I found [this forum post](https://forums.swift.org/t/import-of-c-module-appears-within-extern-c-language-linkage-specification/65606/3) from one of the Apple Swift-C++ developers, made the suggested change, and my projects built again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 ## Main
 
+##### Breaking
+
+* None.
+
+##### Enhancements
+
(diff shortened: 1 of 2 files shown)


johnfairh:
Fixed in #472!

PR #470, opened 2026-01-23
Add WASI platform support for DBL_DECIMAL_DIG

foscomputerservices:
## Summary
- Add WASI platform support for DBL_DECIMAL_DIG to fix compilation on SwiftWasm
- Use conditional compilation to handle platform-specific imports
- Define DBL_DECIMAL_DIG as 17 for WASI (IEEE 754 double precision standard)

## Test plan
- Verify compilation works on SwiftWasm/WASI platform
- Ensure existing Darwin and Glibc platforms continue to work correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Sources/Yams/Representer.swift
vliegeois, line 1: I do not think import Darwin, Glibc ... are necessary.
But indeed, DBL_DECIMAL_DIG is not defined in WASI

So this is enough:
```swift
#if os(WASI)
// WASI/WASM environment - DBL_DECIMAL_DIG not available from system headers
// DBL_DECIMAL_DIG is 17 for IEEE 754 double precision (53-bit mantissa)
    private let DBL_DECIMAL_DIG = 17
#endif
```

Note that I currently have an issue with swift 6.3 and Musl (static compilation --swift-sdk x86_64-swift-linux-musl):
ambiguous use of 'DBL_DECIMAL_DIG'
_float.DBL_DECIMAL_DIG:1:12: note: found this candidate in module '_float'
1 | public var DBL_DECIMAL_DIG: Int32 { get }
  |            `- note: found this candidate in module '_float'

_Builtin_float.DBL_DECIMAL_DIG:1:12: note: found this candidate in module '_Builtin_float'
1 | public var DBL_DECIMAL_DIG: Int32 { get }
  |            `- note: found this candidate in module '_Builtin_float'


Maybe it is an issue in MUSL and should be fixed there

Regards

Vincent

foscomputerservices:
@vliegeois Thank you for the suggested improvements, I have made them as you requested.  I am not very familiar with Musl, but here is what I was able to determine from research:

 - When compiling Swift with the Musl static Linux SDK (x86_64-swift-linux-musl), DBL_DECIMAL_DIG is defined in two different C header modules (_float and _Builtin_float), and Swift can't decide which one to use.             
  - This is a Swift toolchain / Musl SDK bug — the symbol shouldn't be exported from two competing modules.
  
  I hope that this helps to address your concerns.  Please let me know if you would like me to make any further changes.
  
  Thank you!

vliegeois (approved):
Indeed there is an issue for the moment with musl export of DBL_DECIMAL_DIG.
 For now, in my local version, if have hardcoded the value DBL_DECIMAL_DIG =17 but it is not ideal.

Regards

foscomputerservices:
Any thoughts as to whether this can be incorporated or suggestions for improvement?  Thank you!

